Issue #7, opened 2023-06-09
Add tests


rhamzeh:
There are currently no tests, we'd like to have a robust set of tests

PR #25, opened 2023-06-28
chore: adding unit test

adriantam:
## Description
Adding unit tests 

## References
Part of https://github.com/openfga/cli/issues/7

## Review Checklist


cmd/models/list.go
```diff
@@ -30,51 +29,72 @@ import (
...
+	pageIndex := 0
+
+	for {
+		options := client.ClientReadAuthorizationModelsOptions{
+			ContinuationToken: continuationToken,
+		}
+
+		response, err := fgaClient.ReadAuthorizationModels(context.Background()).Options(options).Execute()
+		if err != nil {
+			fmt.Printf("XXX Failed to list models due to %v", err)
```
rhamzeh, line 47: XXX?

cmd/models/write.go
```diff
@@ -19,43 +19,61 @@ import (
...
-		err = json.Unmarshal([]byte(args[0]), &body)
-		if err != nil {
-			fmt.Printf("Failed to parse model due to %v", err)
-			os.Exit(1)
-		}
-		model, err := fgaClient.WriteAuthorizationModel(context.Background()).Body(*body).Execute()
-		if err != nil {
-			fmt.Printf("Failed to write model due to %v", err)
-			os.Exit(1)
+
```
rhamzeh, line 66: If we are returning the error, we should not be printing it (same elsewhere)

mocks/client.go
```diff
@@ -0,0 +1,2734 @@
+// Code generated by MockGen. DO NOT EDIT.
+// Source: client/client.go
+
+// Package mock_client is a generated GoMock package.
+package mock_client
```
rhamzeh, line 5: I know we agreed on having the mocks here, but on seeing them I'm having second thoughts. It may be better for us to move them to the SDK?

Will adding them there require any extra dependencies?

adriantam: Ideally, mockgen can allow us to specify external package (in this case the SDK) and we can generate the mock package as part of make test (and not push it the generated code in).  However, we are not able to do that :(

cmd/models/list.go
adriantam: cleaned up

cmd/models/write.go
adriantam: removed

rhamzeh (approved):

PR #41, opened 2023-07-06
chore(docs): add Read Latest Authorization Model section to the README

aaguiarz:
Added documentation for getting the latest authorization model

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,6 +246,31 @@ fga models **get**
 }
 ```
 
+##### Read the Latest Authorization Model 
+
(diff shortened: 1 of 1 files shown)


rhamzeh (approved):

rhamzeh (commented):
I like this because it is consistent with the SDK docs, but do you think merging this with the upper section makes sense?

aaguiarz:
I'm not sure. I think it's a feature people would want and they are more likely to find it if it's explicitly included with its own section than as a note in the `get` command.

PR #128, opened 2023-08-09
chore: fixed --file examples

aaguiarz:
## Description

Fixed examples where we mentioned you had to write `--file:filename` instead of `--file filename`


rhamzeh:
`--file=<filename>` and `--file <filename>` should both work 🤔

rhamzeh (approved):

PR #168, opened 2023-09-27
feat: create the --simple-json flag for tuple read

### LINKED ISSUE #111: Provide an option in `tuple read` that outputs the format used in `tuple import`
The `tuple read` command displays tuples with this format

```{
      "key": {
        "object":"team:acme-marketing",
        "relation":"member",
        "user":"user:john"
      },
      "timestamp":"2023-07-28T23:06:50.555898Z"
    }
```
However, the `tuple import` command expects them in this format:
```
{
  "user": "user:anne",
  "relation": "can_view",
  "object": "document:roadmap"
}
```

We should have a way to output the format needed by `tuple read`.

It seems that the format used by `tuple import` is simpler to read, so it could make sense to use that one as the default for `tuple read`. However, given that is not the API's output, it will be more expensive to generate and will also be inconsistent with the output results. For that reason, it might be preferred to have a flag to output the simpler format.

I can't thing of a good flag name, some options:

--simple-json
--import-format

gabrielbussolo:
aims to close #111 

I made the read function to return a "complete" version and a "simple" version of the JSON that can be imported by `tuple write` and `tuple delete` with the flag `--file`.

## Description

now if its informed the `--simple-json` flag the response is this:
```
./dist/fga tuple read --store-id 01HB3AZ6RAAGHJQ9HKXV9C89SE --simple-json
[
  {
    "object":"document:Z",
    "relation":"reader",
    "user":"user:gabriel"
  },
  {
    "object":"document:Z",
    "relation":"reader",
    "user":"user:anne"
  },
  {
    "object":"document:Z",
    "relation":"reader",
    "user":"user:jones"
  }
]
```

without the flag:
```
./dist/fga tuple read --store-id 01HB3AZ6RAAGHJQ9HKXV9C89SE              
{
  "tuples": [
    {
      "key": {
        "object":"document:Z",
        "relation":"reader",
        "user":"user:gabriel"
      },
      "timestamp":"2023-09-27T10:58:46.209655Z"
    },
    {
      "key": {
        "object":"document:Z",
        "relation":"reader",
        "user":"user:anne"
      },
      "timestamp":"2023-09-27T10:58:46.208397Z"
    },
    {
      "key": {
        "object":"document:Z",
        "relation":"reader",
        "user":"user:jones"
      },
      "timestamp":"2023-09-27T10:58:46.209576Z"
    }
  ]
}
```

now its possible to use like:
```
 ./dist/fga tuple read --store-id 01HB3AZ6RAAGHJQ9HKXV9C89SE --simple-json > tuple-simple.json
 ./dist/fga tuple delete --store-id 01HB3AZ6RAAGHJQ9HKXV9C89SE --file tuple-simple.json
{
  "successful": [
    {
      "object":"document:Z",
      "relation":"reader",
      "user":"user:gabriel"
    },
    {
      "object":"document:Z",
      "relation":"reader",
      "user":"user:anne"
    },
    {
      "object":"document:Z",
      "relation":"reader",
      "user":"user:jones"
    }
  ],
  "failed":null
}
```

## Review Checklist

### DIFF
--- a/cmd/tuple/read.go
+++ b/cmd/tuple/read.go
@@ -30,8 +30,13 @@ import (
 // MaxReadPagesLength Limit the tuples so that we are not paginating indefinitely.
 var MaxReadPagesLength = 20
 
+type readResponse struct {
+	complete *openfga.ReadResponse
+	simple   []openfga.TupleKey
+}
+
 func read(fgaClient client.SdkClient, user string, relation string, object string, maxPages int) (
-	*openfga.ReadResponse, error,
+	*readResponse, error,
 ) {
 	body := &client.ClientReadRequest{}
 	if user != "" {
@@ -47,7 +52,6 @@ func read(fgaClient client.SdkClient, user string, relation string, object strin
 	}
 
 	tuples := make([]openfga.Tuple, 0)
-
 	continuationToken := ""
 	pageIndex := 0
 	options := client.ClientReadOptions{}
@@ -70,7 +74,14 @@ func read(fgaClient client.SdkClient, user string, relation string, object strin
 		continuationToken = *response.ContinuationToken
 	}
 
-	return &openfga.ReadResponse{Tuples: &tuples}, nil
+	justKeys := make([]openfga.TupleKey, 0)
+	for _, tuple := range tuples {
+		justKeys = append(justKeys, *tuple.Key)
+	}
+
+	res := readResponse{complete: &openfga.ReadResponse{Tuples: &tuples}, simple: justKeys}
+
+	return &res, nil
 }
 
 // readCmd represents the read command.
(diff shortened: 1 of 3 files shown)


cmd/tuple/read.go
```diff
@@ -70,7 +74,14 @@ func read(fgaClient client.SdkClient, user string, relation string, object strin
 		continuationToken = *response.ContinuationToken
 	}
 
-	return &openfga.ReadResponse{Tuples: &tuples}, nil
+	justKeys := make([]openfga.TupleKey, 0)
+	for _, tuple := range tuples {
```
rhamzeh, line 78: Instead of creating a new response, maybe we can just keep returning `*openfga.ReadResponse` as before, and on display just display `*reponse.Tuples` for the simple version

gabrielbussolo: Well, I thought on this, but then later I noticed that we don't have any logic inside of our commands. Normally we are externalizing it on a function. That's why I changed the type instead.

gabrielbussolo: Should I put the for on the command still? 😸

rhamzeh: Sorry for the late reply @gabrielbussolo!

Will merge this as is for now, and will do some cleanup before releasing (we're thinking we want to change the flag to something like `--simple` so that it can be used regardless of the output format).

But I don't want to delay this PR further as we discuss that.

Thanks for taking the initiative on this! 💪🏽 🎉

PR #230, opened 2024-01-17
feat(tuple import): csv optional fields and support any columns order

### LINKED ISSUE #206: Import tuples from a csv
Hello,

We have a private tool to massively import tuples for a given relation from a csv file.

Our use case: 
When migrating permissions from a legacy system to openfga, we need to import all existing permissions as openfga relations. Legacy permissions are exported from the system database as a csv file.
The csv is expected to have 2 columns, the first being the user id and the second the object id. The command has mandatory options to specify the user type, the relation and the object type (In addition to common target server/auth/store related options).

Example
```
cli import-tuples <common options> --user-type user --relation administrator --object-type organization organizations_administrators.csv
```

The command batches the tuples import  (batch size is configurable with an option). 
All failling batches imports are written to a "failed" csv having the same structure as the input one (so it can be re-used as input itself).

Is it something you would be interested in?

Regards,
Yann

le-yams:
## Description
In csv tuple import, making the following fields optional:
 * `user_relation`
 * `condition_name`
 * `condition_context`

Also it supports any columns ordering now.

## References
closes #206 

## Review Checklist


le-yams:
I opened the PR as a draft as it's missing a check for the `condition_name` and `condition_context` (can't defined only one of them, both must be present or absent).

le-yams:
> I opened the PR as a draft as it's missing a check for the `condition_name` and `condition_context` (can't defined only one of them, both must be present or absent).

Thinking about it, is it really true?
The condition name becomes mandatory if the the condition_context header is present but the condition context can be empty thus the field is optional even if the condition_name header is present isn't it?

le-yams:
> The condition name becomes mandatory if the the condition_context header is present but the condition context can be empty thus the field is optional even if the condition_name header is present isn't it?

I've added a commit implementing this behavior

rhamzeh (dismissed):
@le-yams looks good to me!

rhamzeh (approved):

le-yams:
@rhamzeh thank you for the review and fixes 😃

ewanharris (approved):

PR #303, opened 2024-04-17
feat: simplify error messages on tuple write

### LINKED ISSUE #255: Simplify error messages on `tuple write`
When you call `tuple write` and there's an error when writing/deleting a tuple, the CLI returns a `failed` array with one entry for each failed tuple and a `reason` field.

The`reason` field has content similar to:

```
"reason":"Write validation error for POST Write with body {\"code\":\"write_failed_due_to_invalid_input\",\"message\":\"cannot write a tuple which already exists: user: 'user:anne', relation: 'member', object: 'organization:acme': invalid write input\"}\n with error code write_failed_due_to_invalid_input error message: cannot write a tuple which already exists: user: 'user:anne', relation: 'member', object: 'organization:acme': invalid write input"
```

The message is repetitive. In this case, it would be easier to read if it just had the error returned by the server:
```
"reason":"error message: cannot write a tuple which already exists: user: 'user:anne', relation: 'member', object: 'organization:acme': invalid write input"
```

shruti2522:
## Description

I have added the function `extractErrorMssg` to extract out the relevant part of the errror message in the `reason` field.

## References

Fixes #255 

## Review Checklist

### DIFF
--- a/cmd/tuple/import.go
+++ b/cmd/tuple/import.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 
 	openfga "github.com/openfga/go-sdk"
 	"github.com/openfga/go-sdk/client"
@@ -81,6 +82,20 @@ func ImportTuples(
 	return &result, nil
 }
 
+func extractErrMssg(err error) string {
+	errorMsg := err.Error()
+	startIndex := strings.Index(errorMsg, "error message:")
+
+	if startIndex == -1 {
+		return errorMsg
+	}
+
+	errorMsg = errorMsg[startIndex:]
+	errorMsg = strings.TrimSpace(errorMsg)
+
+	return errorMsg
+}
+
 func processWrites(
 	writes []client.ClientWriteRequestWriteResponse,
 ) ([]client.ClientTupleKey, []failedWriteResponse) {
@@ -93,9 +108,10 @@ func processWrites(
 		if write.Status == client.SUCCESS {
 			successfulWrites = append(successfulWrites, write.TupleKey)
 		} else {
+			reason := extractErrMssg(write.Error)
 			failedWrites = append(failedWrites, failedWriteResponse{
 				TupleKey: write.TupleKey,
-				Reason:   write.Error.Error(),
+				Reason:   reason,
 			})
 		}
 	}
@@ -121,9 +137,10 @@ func processDeletes(
 		if delete.Status == client.SUCCESS {
 			successfulDeletes = append(successfulDeletes, deletedTupleKey)
 		} else {
+			reason := extractErrMssg(delete.Error)
 			failedDeletes = append(failedDeletes, failedWriteResponse{
 				TupleKey: deletedTupleKey,
-				Reason:   delete.Error.Error(),
+				Reason:   reason,
 			})
 		}
 	}


cmd/tuple/import.go
```diff
@@ -81,6 +82,15 @@ func ImportTuples(
 	return &result, nil
 }
 
+// function to extract the simplified error message.
+func extractErrorMssg(fullErrorMssg string) string {
+	parts := strings.Split(fullErrorMssg, ": ")
```
ewanharris, line 87: I'm concerned that we're now reducing the error message too much, for example when I write a tuple that already exists then I get `"reason":"invalid write input"` which isn't really an actionable message.

That said, I don't see anything from the Go SDK that would really allow us to improve this as the underlying response from the API isn't available via the SDK unfortunately.

It's not ideal, but if we split on `error message: ` instead then we get two parts and the latter of those two parts seems more suitable `"reason":"cannot write a tuple which already exists: user: 'user:beth', relation: 'viewer', object: 'folder:product-2021': invalid write input"`, what do you think @shruti2522?

ewanharris (changes requested):

cmd/tuple/import.go
shruti2522: I am getting the following output, and for this message the function extracts the last relevant part.
```
shruti@fedora:~/cncf/cli$ ./dist/fga tuple write --store-id=01H0H015178Y2V4CX10C2KGHF4 --file example/folder-document-access_tuples.json
{
  "successful":null,
  "failed": [
    {
      "tuple_key": {
        "object":"folder:product",
        "relation":"owner",
        "user":"user:anne"
      },
      "reason":"No authorization models found for store '01H0H015178Y2V4CX10C2KGHF4'"
    },
    {
      "tuple_key": {
        "object":"folder:product-2021",
        "relation":"parent",
        "user":"folder:product"
      },
      "reason":"No authorization models found for store '01H0H015178Y2V4CX10C2KGHF4'"
    },
    {
      "tuple_key": {
        "object":"folder:product-2021",
        "relation":"viewer",
        "user":"user:beth"
      },
      "reason":"No authorization models found for store '01H0H015178Y2V4CX10C2KGHF4'"
    }
  ]
}

```

ewanharris: Here's the steps I performed in order to get that: result

1. Started an OpenFGA server
2. Created a store `fga store create --name test`
3. Wrote an example model from this repo `fga model write --file ./example/model.fga --store-id <store-id>`
4.  Wrote the initial tuples using `fga tuple write --file ./example/folder-document-access_tuples.json --store-id <store-id>`
5. Wrote the tuples again using `fga tuple write --file ./example/folder-document-access_tuples.json --store-id <store-id>`

Writing the tuples again in step 5 does return output that can be mapped back to the OpenFGA errors (if a developers knows them), but is missing crucial information returned from the server because the error message itself contains the `:` character.

[Looking at the go-sdk](https://github.com/openfga/go-sdk/blob/main/api_open_fga.go#L4286) we can see that the error message returned from the OpenFGA server is appended to the string following `error message: `, below is the output of `write.Error.Error()` for the above test case and the no model found error that demonstrate this. Given that we don't currently have a way to extract that error message from a go-sdk error, I think that this is the safest way to get the error message without losing any extra information.

```
"Write validation error for POST Write with body {\"code\":\"write_failed_due_to_invalid_input\",\"message\":\"cannot write a tuple which already exists: user: 'user:anne', relation: 'owner', object: 'folder:product': invalid write input\"}\n with error code write_failed_due_to_invalid_input error message: cannot write a tuple which already exists: user: 'user:anne', relation: 'owner', object: 'folder:product': invalid write input"
```

```
"Write validation error for POST Write with body {\"code\":\"latest_authorization_model_not_found\",\"message\":\"No authorization models found for store '01HW300QY4HWVEE48WZ1HAK1JJ'\"}\n with error code latest_authorization_model_not_found error message: No authorization models found for store '01HW300QY4HWVEE48WZ1HAK1JJ'"
```

PR #306, opened 2024-04-23
feat: add export store command

### LINKED ISSUE #180: Support exporting store data into the store file format
Add a new `export` command under store: `fga store export`

e.g.
```shell
fga store export --output-file store.fga.yaml
```

This method should:
1. Get the store specified in order to get the name. If this fails for any reason, use the Store ID for the name (Store ID is required)
2. Get the model specified, if not specified get the latest model (Model ID is optional)
3. Read the tuples (all the tuples)
4. Read the assertions for the model specified or latest
5. display that data in the store file file format

This method in essence wraps:
* `store get`
* `model get`
* `tuple read`
* And maps assertions to check test

The data should be displayed in the following format:
```yaml
name: Store Name
model: |
  The model
tuples:
  - user: ...
    relation: ...
    object: ...
tests:
  - name: test 1
    checks:
      - user: ...
        object: ...
        assertions:
          relation1: true
          relation2: false
```

stevehobbsdev:
## Description

Adds a new command to export a store to YAML, either to a file on disk or to STDOUT. The command can write either to STDOUT, or to a file depending on whether the `output-file` flag is specified. If writing to a file, and the file already exists, the user will be prompted to overwrite.

Example help:

```
fga store export --help

Export a store to YAML

Usage:
  fga store export [flags]

Examples:
fga store export

Flags:
  -h, --help                 help for export
      --model-id string      Authorization Model ID
      --output-file string   name of the file to export the store to
      --store-id string      store ID
```

Example output:

```
fga store export --model-id 01HSX7MS9Y4K5M50HBA3HVA07X

name: Test
model: |+
    model
      schema 1.1

    type user

    type group
      relations
        define member: [user]
        define moderator: [user]

tuples:
    - user: user:1
      relation: member
      object: group:admins
    - user: user:1
      relation: member
      object: group:employees
    - user: user:2
      relation: member
      object: group:employees
    - user: user:1
      relation: moderator
      object: group:employees
tests:
    - name: Tests
      check:
        - user: user:1
          object: group:admins
          assertions:
            member: true
        - user: user:2
          object: group:admins
          assertions:
            member: false
        - user: user:1
          object: group:employees
          assertions:
            member: true
            moderator: true
        - user: user:2
          object: group:employees
          assertions:
            member: true
            moderator: false
```

Other notes:

* The code for getting a model, and getting tuples, were refactored into a shared place

## References

Fixes #180 

## Review Checklist
- [X] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [X] The correct base branch is being used, if not `main`
- [X] I have added tests to validate that the change in functionality is working as expected

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,6 +238,72 @@ fga store **import**
 {}
 ```
 
+##### Export Store
+
+###### Command
+fga store **export**
+
+###### Parameters
+* `--store-id`: Specifies the store to export
+* `--output-file`: The file to output the store to (optional, writes to the terminal if omitted)
+* `--model-id`: Specifies the model to export (optional, exports the latest model if omitted)
+* `--max-tuples`: Specifies the max number of tuples to include in the output (option, defaults to 100)
+
+###### Example
+`fga store export --store-id=01H0H015178Y2V4CX10C2KGHF4`
+
+###### Response
+```yaml
+name: Test
+model: |+
+    model
+      schema 1.1
+
+    type user
+
+    type group
+      relations
+        define member: [user]
+        define moderator: [user]
+
+tuples:
+    - user: user:1
+      relation: member
+      object: group:admins
+    - user: user:1
+      relation: member
+      object: group:employees
+    - user: user:2
+      relation: member
+      object: group:employees
+    - user: user:1
+      relation: moderator
+      object: group:employees
+tests:
+    - name: Tests
+      check:
+        - user: user:1
+          object: group:admins
+          assertions:
+            member: true
+        - user: user:2
+          object: group:admins
+          assertions:
+            member: false
+        - user: user:1
+          object: group:employees
+          assertions:
+            member: true
+            moderator: true
+        - user: user:2
+          object: group:employees
+          assertions:
+            member: true
+            moderator: false
+```
+
+If using `output-file`, the response will be written to the specified file on disk. If the desired file already exists, you will be prompted to overwrite the file.
+
 ##### List Stores
 
 ###### Command
--- a/cmd/store/export.go
+++ b/cmd/store/export.go
@@ -1,0 +1,201 @@
+/*
+Copyright © 2023 OpenFGA
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package store
+
+import (
+	"context"
+	"fmt"
+	"math"
+	"os"
+
+	"github.com/openfga/go-sdk/client"
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
+
+	"github.com/openfga/cli/internal/authorizationmodel"
+	"github.com/openfga/cli/internal/cmdutils"
+	"github.com/openfga/cli/internal/confirmation"
+	"github.com/openfga/cli/internal/fga"
+	"github.com/openfga/cli/internal/output"
+	"github.com/openfga/cli/internal/storetest"
+	"github.com/openfga/cli/internal/tuple"
+)
+
+// defaultTuplePageSize defines the default number of pages to return when calling Read.
+const (
+	defaultMaxTupleCount = 100
+)
+
+// buildStoreData compiles all the data necessary to output to file or stdout,
+// or returns an error if this was not successful.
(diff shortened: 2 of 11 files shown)


cmd/store/export_test.go
```diff
@@ -0,0 +1,268 @@
...
+					found = true
+					break
+				}
+			}
+
+			if !found {
+				t.Errorf("Expected check %v not found in output", expectedCheck)
+			}
+		}
+	}
```
stevehobbsdev, line 267: I couldn't just use `DeepEqual` for this as the order of the items in the arrays coming back was not guaranteed and would sometimes be different. Maybe there's a more efficient way to write this though using maps.

cmd/store/export.go
```diff
@@ -0,0 +1,169 @@
...
+				return nil
+			}
+
+			err = os.WriteFile(fileName, storeYaml, 0666)
+			if err != nil {
+				return err
+			}
+		}
+
+		return output.Display(output.EmptyStruct{})
```
stevehobbsdev, line 155: Is this the correct response to return here when writing to a file?

cmd/store/export.go
```diff
@@ -0,0 +1,172 @@
...
+	authModel := authorizationmodel.AuthzModel{}
+	authModel.Set(*model.AuthorizationModel)
+
+	dsl, err := authModel.DisplayAsDSL([]string{"model"})
+	if err != nil {
+		return nil, fmt.Errorf("unable to get model dsl: %w", err)
+	}
+
+	// get the tuples
+	rawTuples, err := tuple.Read(fgaClient, &client.ClientReadRequest{}, 10) //nolint:gomnd
```
stevehobbsdev, line 62: I've ignored it here, but wanted to highlight this magic number for "pages of tuples", is this number appropriate? Do we want to be able to override it with a flag, e.g. `--max-tuples` or similar?

cmd/store/export.go
ewanharris: AFAIK we don't really have a guideline for output (i.e. should it always be in a machine parseable format like json or can we output normal text). I would maybe tend to output a confirmation like `wrote store to file <filename>` but maybe we can revisit that alongside general output?

cmd/store/export.go
stevehobbsdev: This has now been changed to accept a `--max-tuples` flag.

cmd/store/export.go
```diff
@@ -0,0 +1,185 @@
...
+			}
+
+			fileName, _ := cmd.Flags().GetString("output-file")
+			if fileName == "" {
+				fmt.Println(string(storeYaml))
+
+				return nil
+			}
+
+			err = os.WriteFile(fileName, storeYaml, 0o600) //nolint:gomnd
```
ewanharris, line 164: Given that `WriteFile` will overwrite any existing files should we perform a check if the file exists and then show a confirmation prompt to avoid any destructive changes?

cmd/store/export_test.go
```diff
@@ -0,0 +1,275 @@
...
+	"testing"
+	"time"
+
+	openfga "github.com/openfga/go-sdk"
+
+	"github.com/openfga/cli/internal/fga"
+	"github.com/openfga/cli/internal/storetest"
+	"github.com/openfga/cli/internal/tuple"
+
+	// "errors".
```
ewanharris, line 17:
```suggestion
```

ewanharris (commented):
Nice work! Just a comment around writing the file being potentially destructive but other than that looks good!

cmd/store/export.go
stevehobbsdev: Added in https://github.com/openfga/cli/pull/306/commits/dc82f9166984daf314dd3ea669c6a63ec739a421

cmd/store/export.go
stevehobbsdev: I added a message here to say it'd been written to a file, seemed more friendly in the light of adding a confirmation prompt.

stevehobbsdev: Although actually it writes the message as well as `{}` - should it just print the message only?

ewanharris: That's fine to me, I think for a command like this it's potentially expected to have multiple logs output throughout the time it runs

README.md
```diff
@@ -238,6 +238,72 @@ fga store **import**
...
+fga store **export**
+
+###### Parameters
+* `--store-id`: Specifies the store to export
+* `--output-file`: The file to output the store to (optional, writes to the terminal if omitted)
+* `--model-id`: Specifies the model to export (optional, exports the latest model if omitted)
+* `--max-tuples`: Specifies the max number of tuples to include in the output (option, defaults to 100)
+
+###### Example
+`fga store export`
```
ewanharris, line 253: Sorry I missed this earlier, as we require a store ID I think it should be in the example command
```suggestion
`fga store export --store-id=01H0H015178Y2V4CX10C2KGHF4`
```

PR #308, opened 2024-04-26
feat(model/transform): support specifying an output format

### LINKED ISSUE #307: add `output-format` parameter to `fga model transform`
We don't have a way to specify the format for `fga model transform`  command, as in the past we only had two options, the DSL / JSON, so we could infer the output based on the input.

Now the input can be a module file, a DSL file or a JSON file. When we specify a module file, we should able to decide if we want a DSL output or a JSON output, e.g.

`fga model transform --file fga.mod --output-format json`

We currently can't transform from `fga.mod` to JSON directly

ewanharris:
## Description

Adds support for a `--output-format` option that allows specifying `fga` or `json` for the transform output, this is optional and will continue to support outputting `fga` for `json` input and `json` for `fga` and `modular` inputs.

It doesn't prevent transforming to the same as the input as while not useful, it will work so shouldn't be an error case.

## References

Closes #307

## Review Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -559,6 +559,8 @@ fga model **transform**
 ###### Parameters
 * `--file`: File containing the authorization model
 * `--input-format`: Authorization model input format. Can be "fga", "json", or "modular". Defaults to the file extension if provided (optional)
+* `--output-format`: Authorization model output format. Can be "fga" or "json". If input format is `json` then it defaults to DSL, if `fga` or `modular` then defaults to `json` (optional)
+
 
 ###### Example
 `fga model transform --file model.json`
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -559,6 +559,8 @@ fga model **transform**
 ###### Parameters
 * `--file`: File containing the authorization model
 * `--input-format`: Authorization model input format. Can be "fga", "json", or "modular". Defaults to the file extension if provided (optional)
+* `--output-format`: Authorization model output format. Can be "fga" or "json". If input format is `json` then it defaults to DSL, if `fga` or `modular` then defaults to `json` (optional)
```
rhamzeh, line 562:
```suggestion
* `--output-format`: Authorization model output format. Can be "fga" or "json". If not specified, it will default to `fga` if the input format is `json` and to `json` otherwise (optional)
```

rhamzeh (approved):

PR #318, opened 2024-05-02
Used file name as default store name

### LINKED ISSUE #316: Use file name as default store name for `fga store import
When the `store.fga.yaml` file does not have the `name` property set, the `fga store import --file <file>.fga.yaml` command returns an error saying 
```
Error: required flag(s) "name" not set
```

If the `name` property is not specified, we should default the name to the file name, e.g. when running:

```
fga store import --file sample.fga.yaml
```

If the `store.fga.yaml` does not have the `name` property set, the store should be created with the `sample` name.

NeerajNagure:
## Description

Used file name as default store name
Fixes #316 
## Review Checklist

### DIFF
--- a/cmd/store/import.go
+++ b/cmd/store/import.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"strings"
 
 	"github.com/openfga/go-sdk/client"
 	"github.com/spf13/cobra"
@@ -41,11 +42,17 @@ func importStore(
 	storeID string,
 	maxTuplesPerWrite int,
 	maxParallelRequests int,
+	fileName string,
 ) (*CreateStoreAndModelResponse, error) {
 	var err error
 	var response *CreateStoreAndModelResponse //nolint:wsl
-	if storeID == "" {                        //nolint:wsl
-		createStoreAndModelResponse, err := CreateStoreWithModel(clientConfig, storeData.Name, storeData.Model, format)
+	if storeID == "" {                        //nolint:wsl,nestif
+		storeDataName := storeData.Name
+		if storeDataName == "" {
+			storeDataName = strings.TrimSuffix(path.Base(fileName), ".fga.yaml")
+		}
+		createStoreAndModelResponse, err := CreateStoreWithModel(clientConfig, storeDataName, //nolint:wsl
+			storeData.Model, format)
 		response = createStoreAndModelResponse
 		if err != nil { //nolint:wsl
 			return nil, err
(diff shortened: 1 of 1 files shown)


cmd/store/import.go
```diff
@@ -44,8 +44,13 @@ func importStore(
 ) (*CreateStoreAndModelResponse, error) {
 	var err error
 	var response *CreateStoreAndModelResponse //nolint:wsl
-	if storeID == "" {                        //nolint:wsl
-		createStoreAndModelResponse, err := CreateStoreWithModel(clientConfig, storeData.Name, storeData.Model, format)
+	if storeID == "" {                        //nolint:wsl,nestif
+		storeDataName := storeData.Name
+		if storeDataName == "" {
+			storeDataName = storeData.Model
```
ewanharris, line 50: I think here we potentially want to use [path.Base](https://pkg.go.dev/path#Base) to get the filename, and then use [strings.TrimSuffix](https://pkg.go.dev/strings#TrimSuffix) to remove `fga.yaml` from the filename which should leave us with the store name to use

ewanharris (changes requested):
Thanks for the PR @NeerajNagure!

It looks like instead this PR is setting the value of the `Model` property, but I think the idea is to use the filename of the provided store file as the store name if `name` is not provided. So if I run `fga store import --file my-store.fga.yaml` then it should be imported as `my-store`.

I've left a comment around how we could potentially achieve this

cmd/store/import.go
NeerajNagure: @ewanharris I have not taken the fileName as an argument in importStore function and extracted the store name part.Am I correct in this approach?

ewanharris: @NeerajNagure I think it's nearly there. Assuming I'm running `fga store import --file /Users/ewan/my_store.fga.yaml` then as we're currently doing `strings.TrimSuffix(fileName, ".fga.yaml")` we will end up with a store name that includes path (i.e. `/Users/ewan/my_store`), whereas we want just `my_store`,

Doing `storeDataName = strings.TrimSuffix(path.Base(fileName), ".fga.yaml")` should allow us to retrieve the filename (`my_store.fga.yaml`)  and then trim `".fga.yaml"` from it to get `my_store`

aaguiarz:
If it helps, we are doing something similar in 

```fga create store --model model.fga```

where use 'model' as the name of the store.

NeerajNagure:
@aaguiarz @ewanharris I have made the required changes.Please review the PR

ewanharris (approved):
This looks great! Thanks @NeerajNagure!

PR #345, opened 2024-06-08
feat(generate): generate pkl testing

toniphan21:
Introduce new command `generate` to generate some code for testing.

## Description
The newly introduce command `generate` has a subcommand `pkl` which generate a small `pkl` code library with assertions and types from given openFGA model. The structure of generated directory looks like:

- `testing/lib/test.pkl`: small library file to match the syntax of pkl to yaml file for testing a model - this file always be overridden when run the command
- `testing/lib/gen.pkl`: generated types and assertions based on model - this file always be overridden when run the command 
- `testing/types.pkl`: a types definitions where we can write some functions to make the tests setup looks nicer - only generated if the file is not found.
- `testing/types.pkl`: assertions library where we can write some functions to make the tests assertions looks nicer - only generated if the file is not found.
- `testing/test_example.pkl`: an example test - only generated if the file is not found.
- `testing/run`: a simple bash script helper to run a test or all tests - only generated if the file is not found.

## Example

Run this command with the model `custom-roles` in [sample-store](https://github.com/openfga/sample-stores/blob/main/stores/custom-roles/model.fga).

```
fga generate pkl  --file=path-to-model/model.fga 
```

The generated contents are:

`testing/run`
```
#!/bin/bash
param=$1
if [ "$param" == "all" ]; then
  tests=`ls ./test*.pkl`
  for test in $tests
  do
    echo "--- running test file: $test"
    pkl eval -f yaml $test | fga model test --tests /dev/stdin
    echo "--- finished"
  done
else
  pkl eval -f yaml $param | fga model test --tests /dev/stdin
fi
```

...

## Review Checklist


ewanharris:
Hey @toniphan21, thanks for this PR and the well thought out description to bring us up to speed! ❤️ 

Just to set some expectations, it might take us a little while to get around to fully reviewing this PR. We'd like to make sure we fully understand the pieces that are new to us (pkl, generation of files outside the model) and how the full model development cycle looks with this change (e.g. local and CI usage, potentially expanding to how this may fit into areas like the playground).

We're definitely excited to see the suggestions you have for improving the testing of a model. Out of curiosity is this something you're using to author tests currently?

toniphan21:
Hey @ewanharris, thank you for reviewing this PR. I understand it's a big one, so I don't expect the review process to be fast. Also, you can see I haven't added any tests yet. The reason is that I want to gather your feedback before adding them.

To answer your question, yes, I do use these tests at work to verify the model with non-technical stakeholders. That's why readability is the main focus.

Let me walk you through the process of how I came up with the idea, and then I'll suggest some approaches for integrating pkl into openfga's ecosystem.

First, I started with the yaml file for testing a model:

```yaml
name: test
model: ''
tuples:
- user: user:1
  relation: member
  object: org:1
tests:
- name: whatever
  check:
  - user: user:1
    object: object
    assertions:
      allow: true
```
Then I wrote a `pkl` file which could be converted to the exact yaml file above:
```pkl
name: String = "test"
model: String = ""
tuples: Listing<Tuple> = new {
  new {
    user = "user:1"
    relation = "member"
    object = "org:1"
  }
}
tests: Listing<TestCase> = new {
  new {
    name = "whatever"
    check {
      new {
        user = "user:1"
        object = "object"
        assertions {
          ["allow"] = true
        }
      }
    }
  }
}

class Test {
  name: String
  model: String?
  tuples: Listing<Tuple>
  tests: Listing<TestCase>
}

class Tuple {
  user: String
  relation: String
  object: String
}

class TestCase {
  name: String?
  check: Listing<Check>
}

class Check {
  user: String
  object: String
  assertions: Mapping<String, Boolean>
}
```

As you can see, the raw pkl file is longer than the yaml file. However, there are some upsides to using pkl:

- I can separate a single pkl file using modules, which means I can import files.
- I can write functions to express business logic in a more readable way. For example:

```pkl
new {
    user = "user:1"
    relation = "member"
    object = "org:1"
  }
```
becomes
```pkl
// somewhere at the top of the pkl file
local Anna = new User("1")
local OpenFGA = new Org("1")

Anna.member(OpenFGA)
```

The benefit is that I can name `member` whatever I want and still produce the same output. I can condense every setup (producing a tuple) into one line. Another advantage is that in the function, we can produce more than one tuple (for example, a user belongs to an org - one tuple, and the org has the member - another tuple).

In short, the `pkl` file is just a nicer way to express ideas or business logic; in the end, it will be converted to yaml and tested using the openfga cli.

However, I faced a problem where I have to write a `pkl` class every time my model changes. That’s why I want to `generate pkl` files automatically based on the model, saving time and avoiding mistakes.

I suggest that:

- Write documentation on how to convert a pkl file to yaml, merge this MR, and add the fga generate pkl command to the documentation to show users how to create YAML files faster with more readable syntax.
- Continue integrating listObject and listUser functionality into the pkl generation command.
- Integrate pkl files into the openfga CLI, which means running pkl directly instead of converting it to yaml (this task isn’t too big because the structure generated from the pkl file is very similar to yaml, even cleaner, and I’m happy to do it if OpenFGA thinks it makes sense).
- Once the fga cli can run pkl directly, we could integrate it into the playground.

I think this is a lot of work, and it’s not easy to integrate a new language into an ecosystem. However, I believe pkl offers a lot compared to yaml, and generating source code from a model is a good start. Please test this PR with any models in the sample store and give me feedback. I hope the idea is clearer after you've tried it.

rhamzeh:
@toniphan21 - thanks for your work on getting this PR!

We discussed it earlier with the team, and the consensus seems to be we'd still rather hold off on it for now, for a few reasons:
1. Maintenance burden
2. Not a lot of folks asking for this specifically

Long term, we believe this is very valuable, but we think it should probably belong as a plugin to the CLI. The problem is that the fga cli is not yet setup to enable plugins, so that work needs to be done first to unblock you.

cc @dyeam0 / @ewanharris

aaguiarz:
I'm closing the PR for the time being

PR #506, opened 2025-06-05
feat: support multiple tuple files

SoulPancake:
I was giving this feature a shot,
I was able to make it work when you provide the source test files like such
```
name: Test Store
model_file: ./test.fga
#tuple_file: ./singletuple.yaml
tuple_files:
  - ./tuple.yaml
  - ./tuple_2.yaml
tests:
  - name:  Admin checks
    description: To test what   authorisation on all objects
    check:
    ```
    
   Although this is still in draft mode so let me know if this is what we expect, I can finish the tests then

## Review Checklist

### DIFF
--- a/internal/storetest/storedata.go
+++ b/internal/storetest/storedata.go
@@ -65,12 +65,13 @@ type ModelTest struct {
 }
 
 type StoreData struct {
-	Name      string                            `json:"name"       yaml:"name"`
-	Model     string                            `json:"model"      yaml:"model"`
-	ModelFile string                            `json:"model_file" yaml:"model_file,omitempty"` //nolint:tagliatelle
-	Tuples    []client.ClientContextualTupleKey `json:"tuples"     yaml:"tuples"`
-	TupleFile string                            `json:"tuple_file" yaml:"tuple_file,omitempty"` //nolint:tagliatelle
-	Tests     []ModelTest                       `json:"tests"      yaml:"tests"`
+	Name       string                            `json:"name"       yaml:"name"`
+	Model      string                            `json:"model"      yaml:"model"`
+	ModelFile  string                            `json:"model_file" yaml:"model_file,omitempty"` //nolint:tagliatelle
+	Tuples     []client.ClientContextualTupleKey `json:"tuples"     yaml:"tuples"`
+	TupleFile  string                            `json:"tuple_file" yaml:"tuple_file,omitempty"` //nolint:tagliatelle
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: internal/storetest/storedata.go [hunk 1]
```diff
@@ -115,6 +116,24 @@ func (storeData *StoreData) LoadTuples(basePath string) error {
 		}
 	}
 
+	if storeData.TupleFiles != nil {
```
SoulPancake, line 119: @rhamzeh Can you take a look at this if possible?

This would be a backwards compatible solution without messing with the way the Tuples are read for a single tuple file

SoulPancake: Currently this would cause an issue if both the tuple file and tuple files array are being passed

SoulPancake: We can choose to directly have the files in the list format inside tupleFiles
and remove separate file/files altogether

aaguiarz:
@SoulPancake yeah, this makes sense.

Please make sure you add tests to make sure the new format and old format work, and update docs

SoulPancake:
@aaguiarz 
What should the expected behaviour be if the tuple file ( single ) and tuple files array both are provided?

SoulPancake:
> @aaguiarz What should the expected behaviour be if the tuple file ( single ) and tuple files array both are provided?

My proposal would be
1. If a tuple file alone is provided, just load those tuples.
2. If multiple tuple files are provided, load all of them
3. If both tuple file and tuple files array is provided, load from all the files but do a duplicate check before loading tuples?

internal/storetest/storedata.go
```diff
@@ -101,17 +102,57 @@ func (storeData *StoreData) LoadModel(basePath string) (authorizationmodel.Model
 	return format, nil
 }
 
+//nolint:gocognit,cyclop
```
SoulPancake, line 105: @aaguiarz 
I am wondering if we do a tuple duplicate check like such
the cyclomatic complexity increases failing these linters
What is your suggestion?

SoulPancake:
>I think it's better to surface that issue to the user than to hide it.
So programmatically we still keep the map? and error out on duplicate tuples.

SoulPancake:
@aaguiarz I have removed dedupe logic and updated this like we discussed, Can you please review this?

SoulPancake:
NP, Sure I'll rebase it on Sunday. Thanks @rhamzeh

rhamzeh:
A review from someone not me is required now that I've modified it. Will ask around :)

PR #664, opened 2026-03-30
test(tuplefile): add unit tests for CSV and JSONL parsing

archy-rock3t-cloud:
Add unit tests for CSV and JSONL 
  parsing

## Summary by CodeRabbit

* **Tests**
  * Added a comprehensive test suite for data parsing covering CSV and JSONL inputs: header validation, required-field checks, various column orders, row extraction, optional fields handling, condition parsing (including JSON context), skipping blank lines, and detailed error cases for invalid or missing data. Includes unit tests for condition-column combinations and header validation.

### DIFF
--- a/internal/tuplefile/tuplefile_test.go
+++ b/internal/tuplefile/tuplefile_test.go
@@ -1,0 +1,450 @@
+package tuplefile
+
+import (
+	"bytes"
+	"encoding/csv"
+	"testing"
+
+	openfga "github.com/openfga/go-sdk"
+	"github.com/openfga/go-sdk/client"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// CSV header parsing tests
+
+func TestReadHeaders_ValidRequiredHeaders(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		name     string
+		csv      string
+		expected csvColumns
+	}{
+		{
+			name: "minimal required headers",
+			csv:  "user_type,user_id,relation,object_type,object_id\n",
+			expected: csvColumns{
+				UserType:         0,
+				UserID:           1,
+				UserRelation:     -1,
+				Relation:         2,
+				ObjectType:       3,
+				ObjectID:         4,
+				ConditionName:    -1,
+				ConditionContext: -1,
+			},
+		},
+		{
+			name: "all headers present",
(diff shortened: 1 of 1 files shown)


archy-rock3t-cloud:
Thanks @Copilot / @coderabbitai — addressed the review:

- **Unused `expected csvColumns` fixture**: now asserted by calling `readHeaders` directly in the success path and comparing the parsed `csvColumns` to `tc.expected`, so header-index mapping regressions are actually caught.
- **Brittle line-number assertion**: softened `"failed to read condition context on line 0"` to a substring check on `"failed to read condition context"` so the test isn't coupled to the parser's 0-based line numbering.
- **`tc := tc` loop-var capture**: tried this first (the classic pre-Go-1.22 fix), but the repo's `golangci-lint` config flags it as `copyloopvar`: *"The copy of the 'for' variable can be deleted (Go 1.22+)"*. Since `go.mod` is at `go 1.25`, per-iteration scoping is the language default and the explicit re-bind would actually fail lint. Renamed `tc` → `testCase` instead, which both satisfies `varnamelen` (the original Lints failure) and makes parallel safety explicit.
- Also fixed the `wsl_v5` whitespace violations that were tripping the Lints job (separator lines between `var` decls and following `err :=` assignments).

Rebased on `main`. Local `go test ./internal/tuplefile/... -race -count=1` passes; `golangci-lint v2.10.1 -c .golangci.yaml ./internal/tuplefile/...` reports `0 issues`.

The remaining `Security Audits` red is `GO-2026-4947` in the Go stdlib (`tls.Conn.Read`) reachable via `bufio.Scanner.Scan` in `read.go` — pre-existing on `main` and being handled separately in `chore/vulncheck-fix`, unrelated to this PR. `License Compliance` is FOSSA infra.

curfew-marathon:
This PR has had no human activity for 90 days, so it has been marked `lifecycle/stale`.

This is automated backlog grooming, not a judgment on the work.

What happens next, unless there is activity:
- In 30 days, it will move to `lifecycle/closing-soon`
- 30 days after that, it will be closed

To keep it open, push a commit or leave a comment, and the clock will reset. For work that should not auto-close, such as an RFC or long-running experiment, ask a maintainer to add `lifecycle/frozen`.

Siddhant-K-code:
Non-blocking: `internal/tuplefile/tuplefile_test.go:413` — use `assert.Equal(t, expected, condition)` instead of field assertions. It removes the duplicate name check and verifies the full condition. No blocking findings.

Issue #669, opened 2026-04-08
Support custom headers


alexanderzobnin:
When openfga is embedded in another application as library, there might be different authentication method in front of GRPC connection. For example, in grafana we use [token-based authentication](https://github.com/grafana/grafana/blob/daded900b5e2b9df0840f75b68b4b777e205978b/pkg/services/authz/zanzana.go#L312) for GRPC connection to the service which embeds openfga. So in order to use openfga-cli we need to pass access token in special header which is not possible currently with CLI.

aaguiarz:
Thanks @alexanderzobnin ! Can you suggest a syntax that would work for you?

alexanderzobnin:
I'm working on PR, so we can discuss it one it's ready.

alexanderzobnin:
I suggested `--custom-headers` flag which allows to pass any required headers, so it's quite flexible.

PR #670, opened 2026-04-08
feat: support custom headers

alexanderzobnin:

Solves https://github.com/openfga/cli/issues/669

#### How is it being solved?

Provide new `--custom-headers` flag which allows to pass custom user-defined headers which are then included to every request to API.

Implemented new `--custom-headers` option by utilizing existing `DefaultHeaders` client config field.

## Review Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for custom HTTP headers via the `--custom-headers` CLI flag. This repeatable flag accepts header values in `name:value` format, enabling users to include custom headers in requests.


internal/fga/fga.go
```diff
@@ -95,6 +96,20 @@ func (c ClientConfig) getClientConfig() *client.ClientConfiguration {
...
+
+func (c ClientConfig) getCustomHeaders() map[string]string {
+	headers := map[string]string{}
+	for _, header := range c.CustomHeaders {
+		parts := strings.SplitN(header, ":", 2)
+		if len(parts) == 2 {
+			head := strings.TrimSpace(parts[0])
+			value := strings.TrimSpace(parts[1])
+			headers[head] = value
+		}
```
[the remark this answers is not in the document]
rhamzeh: @alexanderzobnin I think it would be good to throw an error here, otherwise folks might not notice that their headers are being ignored

README.md
```diff
@@ -151,6 +151,7 @@ For any command that interacts with an OpenFGA server, these configuration value
 | Token Audience         | `--api-audience`     | `FGA_API_AUDIENCE`     | `api-audience`     |
 | Store ID               | `--store-id`         | `FGA_STORE_ID`         | `store-id`         |
 | Authorization Model ID | `--model-id`         | `FGA_MODEL_ID`         | `model-id`         |
+| Custom Headers         | `--custom-headers`   | `FGA_CUSTOM_HEADERS`   | `custom-headers`   |
```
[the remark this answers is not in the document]
rhamzeh: I kind of agree that we need better docs than this for folks to know how to use it.
e.g. something like the section below just above the **Commands** header:

```
#### Custom Headers

You can add custom HTTP headers to all requests sent to the API using the `--custom-headers` flag. Headers are specified in `<name>: <value>` format, and the flag can be repeated to add multiple headers.

##### Flag
\```shell
--custom-headers "Header-Name: header-value"
\```

##### Example
\```shell
fga store list --custom-headers "X-Custom-Header: value1" --custom-headers "X-Request-ID: abc123"
\```

##### Configuration

Custom headers can also be configured via the CLI environment variable or the configuration file:

| Name           | Flag                 | CLI                    | ~/.fga.yaml        |
|----------------|----------------------|------------------------|---------------------|
| Custom Headers | `--custom-headers`   | `FGA_CUSTOM_HEADERS`   | `custom-headers`    |

Example `~/.fga.yaml`:
\```yaml
api-url: https://api.fga.example
store-id: 01H0H015178Y2V4CX10C2KGHF4
custom-headers:
  - "X-Custom-Header: value1"
  - "X-Request-ID: abc123"
\```
```

rhamzeh (changes requested):

rhamzeh:
One more thing - can we add some tests for these, just to make sure all is good? And please make sure to run `make lint` and `make format` 🙏🏽

emilic:
Just a general comment, i happened to look at this and noticed that for env variables we have a bug in:

```
func BindViperToFlags(cmd *cobra.Command, viperInstance *viper.Viper) {
	cmd.Flags().VisitAll(func(flag *pflag.Flag) {
		configName := flag.Name

		if !flag.Changed && viperInstance.IsSet(configName) {
			value := viperInstance.Get(configName)
			err := cmd.Flags().Set(flag.Name, fmt.Sprintf("%v", value)) <---- Here
			cobra.CheckErr(err)
		}
	})

	for _, subcmd := range cmd.Commands() {
		BindViperToFlags(subcmd, viperInstance)
	}
}
```

If the custom headers is passed as an env variable it will stringify the the string array.  It will probably work with one entry.  

not sure how likely env vars are in your use case but i assume it would be for some auth headers and maybe not multiple.

this is not a problem with your PR but its a bug that came to light because of it.  @rhamzeh - ill create a separate issue for this.

UPDATE: and i just noticed that co-pilot already flagged this.

rhamzeh (dismissed):

rhamzeh:
Thanks @alexanderzobnin - we'll fix the lint issue and merge :)

@emilic please do!

rhamzeh:
@alexanderzobnin can I ask you to either:

* allow write access to your branch to maintainers so we can fix the lint issues (See https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)
* or fix the lint issues and push them on your end?

These are the changes needed:
* Lint (instead of applying, just run make lint on your end): [0001-chore-fix-lint.patch](https://github.com/user-attachments/files/26746364/0001-chore-fix-lint.patch)
* Fixes the security failure: [0002-chore-bump-toolchain-to-1.26.2.patch](https://github.com/user-attachments/files/26746366/0002-chore-bump-toolchain-to-1.26.2.patch)

internal/fga/fga_test.go
```diff
@@ -0,0 +1,183 @@
...
+			t.Parallel()
+
+			var capturedHeaders http.Header
+
+			server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				capturedHeaders = r.Header.Clone()
+				w.Header().Set("Content-Type", "application/json")
+				w.WriteHeader(http.StatusOK)
+				_, _ = w.Write([]byte(`{"stores": []}`))
+			}))
```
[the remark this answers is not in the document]
senojj: Yes this should be addressed before merging.

internal/cmdutils/bind-viper-to-flags.go
```diff
@@ -31,12 +31,30 @@ func BindViperToFlags(cmd *cobra.Command, viperInstance *viper.Viper) {
...
 }
+
+// viperValueToStrings converts a Viper config value to a slice of strings
+// suitable for pflag.Set calls. Slice values (from YAML lists) produce one
+// string per element; scalar values produce a single-element slice.
+func viperValueToStrings(value any) []string {
+	sliceValue, ok := value.([]any)
+	if !ok {
+		return []string{fmt.Sprintf("%v", value)}
+	}
```
senojj: This will be false for every kind of slice that is not `[]any`. For instance `[]string` or `[]int`. Use the reflect package to identify a slice of any kind, unless `(*viper.Viper).Get()` is always guaranteed to return `[]any` for any slice -- I do not believe it is.

internal/fga/fga.go
```diff
@@ -95,6 +113,27 @@ func (c ClientConfig) getClientConfig() *client.ClientConfiguration {
...
+}
+
+func (c ClientConfig) getCustomHeaders() (map[string]string, error) {
+	headers := map[string]string{}
+
+	for _, header := range c.CustomHeaders {
+		parts := strings.SplitN(header, ":", 2)
+		if len(parts) != 2 {
+			return nil, fmt.Errorf("invalid custom header %q: %w", header, ErrInvalidHeaderFormat)
+		}
```
senojj: use `strings.Cut` instead, as it is faster and does not need to allocate.

```go
name, value, _ := strings.Cut(header, ":")
name, value = strings.TrimSpace(name), strings.TrimSpace(value)
if name == "" {
    return nil, fmt.Errorf("...")
}
headers[name] = value
```

internal/fga/fga.go
```diff
@@ -95,6 +113,27 @@ func (c ClientConfig) getClientConfig() *client.ClientConfiguration {
...
 			MinWaitInMs: MinSdkWaitInMs,
 		},
-		Debug: c.Debug,
+		Debug:          c.Debug,
+		DefaultHeaders: customHeaders,
+	}, nil
+}
+
+func (c ClientConfig) getCustomHeaders() (map[string]string, error) {
+	headers := map[string]string{}
```
senojj: prevent unnecessary allocations:
```go
header := make(map[string]string, len(c.CustomHeaders))
```

internal/requests/rampup_test.go
```diff
@@ -47,15 +47,15 @@ func TestRampUpAPIRequests_RampUpRate(t *testing.T) {
...
 		mutex     sync.Mutex
 	)
 
 	requestsList := make([]func() error, 10)
 	for i := range requestsList {
 		requestsList[i] = func() error {
 			mutex.Lock()
-			atomic.AddInt32(&callCount, 1)
+			callCount.Add(1)
 			mutex.Unlock()
```
senojj, line 59: I'm more curious as to why a mutex is used here.....

.mise.toml
senojj, line 1: ???

alexanderzobnin: This is what @rhamzeh asked to do (this was a part of go toolchain bump).

internal/requests/rampup_test.go
alexanderzobnin: I have no idea, this one is change produced by `make lint`.

.mise.toml
alexanderzobnin: @rhamzeh should we keep it?

internal/fga/fga.go
```diff
@@ -95,6 +113,28 @@ func (c ClientConfig) getClientConfig() *client.ClientConfiguration {
...
+
+	for _, header := range c.CustomHeaders {
+		name, value, found := strings.Cut(header, ":")
+		name, value = strings.TrimSpace(name), strings.TrimSpace(value)
+		if !found {
+			return nil, fmt.Errorf("invalid custom header %q: %w", header, ErrInvalidHeaderFormat)
+		}
+		if name == "" && value == "" {
+			return nil, fmt.Errorf("invalid custom header %q: %w", header, ErrInvalidHeaderFormat)
+		}
```
senojj: An empty string is a valid value for a header, so both of these checks can likely be elided. The `!found` check _could_ be reasonable if you want to be super strict about the existence of a ':'.

internal/cmdutils/bind-viper-to-flags.go
```diff
@@ -31,12 +32,30 @@ func BindViperToFlags(cmd *cobra.Command, viperInstance *viper.Viper) {
...
 }
+
+// viperValueToStrings converts a Viper config value to a slice of strings
+// suitable for pflag.Set calls. Slice values (from YAML lists) produce one
+// string per element; scalar values produce a single-element slice.
+func viperValueToStrings(value any) []string {
+	rv := reflect.ValueOf(value)
+	if rv.Kind() != reflect.Slice {
+		return []string{fmt.Sprintf("%v", value)}
+	}
```
senojj: while handling for slices, wouldn't hurt to handle for arrays, just in case:
```go
select rv.Kind() {
case reflect.Slice, reflect.Array:
default:
    return []string{fmt.Sprintf("%v", value)}
}

// ... the rest removed for brevity
```

internal/fga/fga.go
alexanderzobnin: Yah, maybe it's too strict. Currently it's not allowed to use `--custom-headers "X-Custom-Header"`, you have to provide `--custom-headers "X-Custom-Header:"`, so maybe I should change it to allow both.

>An empty string is a valid value for a header

That's true, but I'm trying to differentiate cases where no header name provided and both header and value are missing (`--custom-headers ""` or `--custom-headers ":"`)

Do you think we can simplify this and return the same error about missing header name?

senojj: yeah, just validating the header field's name should be sufficient:
```go
if name == "" {
   // return error about invalid header format
}
```
that will prevent cases like `--custom-headers ""` and `--custom-headers ":"`.

you can entirely elide the checks on the header field value, because it is fine for it to be empty (e.g. `--custom-headers "X-Custom-Header:"`).

internal/cmdutils/bind-viper-to-flags.go
alexanderzobnin: Sorry, using `switch case` causes this linter error, so I used `if {}`:
```
INFO [runner] linters took 1.945138333s with stages: goanalysis_metalinter: 1.934925791s
internal/cmdutils/bind-viper-to-flags.go:52:2: missing cases in switch of type reflect.Kind: reflect.Invalid, reflect.Bool, 
reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64, reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Uintptr, reflect.Float32, reflect.Float64, reflect.Complex64, 
reflect.Complex128, reflect.Chan, reflect.Func, reflect.Interface, reflect.Map, reflect.Pointer|reflect.Ptr, reflect.String, reflect.Struct, reflect.UnsafePointer (exhaustive)
        switch reflectValue.Kind() {
        ^
```

senojj: Haha, darn linter. No problem :)

.mise.toml
senojj: @rhamzeh was the addition of this file the intent? I wasn't sure if we were actually using the tool in the project.

emilic:
Here's a fix for multiple values via env vars (#678) — feel free to merge into your branch, or I can merge it separately after yours.

```go
if strings.HasPrefix(str, "[") && strings.HasSuffix(str, "]") {
    var parsed []string
    if err := json.Unmarshal([]byte(str), &parsed); err == nil {
        return parsed
    }
}
```

This would go in `viperValueToStrings` before the scalar fallback, allowing:
```sh
FGA_CUSTOM_HEADERS='["X-Foo: bar","X-Baz: qux"]' fga store list
```

internal/cmdutils/bind-viper-to-flags.go
```diff
@@ -31,12 +32,31 @@ func BindViperToFlags(cmd *cobra.Command, viperInstance *viper.Viper) {
...
+
+	result := make([]string, 0, reflectValue.Len())
+	for i := range reflectValue.Len() {
+		result = append(result, fmt.Sprintf("%v", reflectValue.Index(i).Interface()))
+	}
+
+	return result
+}
```
senojj, line 62: I was doing some reading on #678 and instead of using `viperInstance.Get(configName)` in `BindViperToFlags`, why don't we use `viperInstance.GetStringSlice(configName)`?

https://pkg.go.dev/github.com/spf13/viper#GetStringSlice

That would be guaranteed to always return a slice of strings, regardless of the original value type. Additionally, `fmt.Sprintf` won't be necessary anymore, as the logic reduces to:
```go
for _, strVal := range viperInstance.GetStringSlice(configName) {
	cobra.CheckErr(cmd.Flags().Set(flag.Name, strVal))
}
```
And the `viperValueToStrings` function can be removed entirely.

emilic: this is fine for non env variables but if you follow the call chain for viperInstance.GetStringSlice(..) it calls cast.ToStringSlice(v.Get(key)) which calls:

```go
// ToStringSliceE casts any value to a []string type.
func ToStringSliceE(i any) ([]string, error) {
	if a, ok, err := toSliceEOk[string](i); ok {
		if err != nil {
			return nil, err
		}

		return a, nil
	}

	var a []string

	switch v := i.(type) {
	case string:
		return strings.Fields(v), nil
	case any:
		str, err := ToStringE(v)
		if err != nil {
			return nil, fmt.Errorf(errorMsg, i, i, a)
		}

		return []string{str}, nil
	default:
		return nil, fmt.Errorf(errorMsg, i, i, a)
	}
}
```

case string: calls strings.Fields(v) which does split on whitespace.  That wouldn't fix it.

senojj: @emilic as you state, comma separated values are not parsed by `GetStringSlice`; it ultimately uses `strings.Fields` which only separates on whitespace.

However, would it not be feasible for us to just document that multiple headers should be separated by whitespace (e.g. "X-Fizz:one X-Buzz:two") in the env var, or do we already document the use of comma separated values?

I also understand the argument of wanting to allow whitespace within header values, but the same can be said of commas and brackets.

emilic: @senojj - I think its fine to just document the approach.  I am not partial to one way or another, every choice has its pros/cons.  What you suggest would solve the problem as long as its documented.

alexanderzobnin:
Hi, folks! Could you clarify what's left on my side? Any changes you want me to make? I see some discussions related to another PR #679 , but not sure what should I do about this.

senojj:
> Hi, folks! Could you clarify what's left on my side? Any changes you want me to make?

It would be more robust to use:
```go
for _, strVal := range viperInstance.GetStringSlice(configName) {
	cobra.CheckErr(cmd.Flags().Set(flag.Name, strVal))
}
```
The `viperValueToStrings` function can be removed entirely.

senojj:
@alexanderzobnin don't remove the BindViperToFlags tests. We should change the assertions to the actual expectations. It looks like the whitespace just needs to be removed from the header field values (e.g. "X-Fizz: buzz" becomes "X-Fizz:buzz")

alexanderzobnin:
@senojj the problem is not assertion, but the import:
```
internal/cmdutils/bind-viper-to-flags_test.go:22:2: import 'github.com/spf13/cobra' is not allowed from list 'test' (depguard)
        "github.com/spf13/cobra"
        ^
internal/cmdutils/bind-viper-to-flags_test.go:23:2: import 'github.com/spf13/viper' is not allowed from list 'test' (depguard)
        "github.com/spf13/viper"
        ^
2 issues:
```

senojj:
> ```
> internal/cmdutils/bind-viper-to-flags_test.go:22:2: import 'github.com/spf13/cobra' is not allowed from list 'test' (depguard)
>         "github.com/spf13/cobra"
>         ^
> internal/cmdutils/bind-viper-to-flags_test.go:23:2: import 'github.com/spf13/viper' is not allowed from list 'test' (depguard)
>         "github.com/spf13/viper"
>         ^
> 2 issues:
> ```

@rhamzeh any reason we are explicitly blocking cobra and viper packages from tests?

alexanderzobnin:
I added packages to allowlist for the linter (tests)

internal/cmdutils/bind-viper-to-flags_test.go
```diff
@@ -0,0 +1,100 @@
...
+		expected []string
+	}{
+		{
+			name: "slice value produces one flag value per element",
+			value: []any{
+				"X-Custom-Header: value1",
+				"X-Request-ID: abc123",
+			},
+			expected: []string{"X-Custom-Header: value1", "X-Request-ID: abc123"},
+		},
```
senojj, line 45: Could you add a test case that would mimic an env var value?
e.g. "X-Custom-Header:value1 X-Request-ID:abc123"

internal/fga/fga.go
```diff
@@ -32,7 +34,11 @@ const (
 	MinSdkWaitInMs = 500
 )
 
-var userAgent = "openfga-cli/" + build.Version
+var (
+	userAgent = "openfga-cli/" + build.Version
+
+	ErrInvalidHeaderFormat = errors.New("expected format \"Header-Name: value\"")
```
senojj: In the expected format, remove the space between the colon ':' and the first character of the field value. This will make the expectation consistent with the env var requirement.

senojj: After this change, everything else looks good 👍

senojj (approved):

PR #704, opened 2026-06-11
refactor: drop gocsv to stdlib encoding/csv 

### LINKED ISSUE #674: fix: remove gocsv dependency
Remove the github.com/gocarina/gocsv dependency by implementing a local CSV marshaler using stdlib encoding/csv and reflect for struct tags.

**Files:** 
- internal/output/marshal.go:77
- cmd/tuple/read.go:50–58

**Changes:**
- Create a local marshalCSV(data any) helper function using encoding/csv and reflect
- Parse csv struct tags to generate headers and rows
- Replace gocsv.MarshalBytes() call with the new helper

SoulPancake:
## Description

 - Replace gocsv.MarshalBytes with a local reflect-based marshalCSV
    helper using encoding/csv, fully removing the gocsv dependency
    (closes #674)
  - Add marshalCSV test coverage (escaping, omitempty, nil pointers,
    non-string fields, error cases)

## Review Checklist

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test suite for CSV output formatting, covering edge cases like escaped characters, multi-row output, and error scenarios.

* **Refactor**
  * Improved internal error handling mechanisms and reduced external dependency requirements for better stability and maintainability.


go.mod
```diff
@@ -49,6 +47,7 @@ require (
 	github.com/grpc-ecosystem/go-grpc-middleware/v2 v2.3.3 // indirect
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.29.0 // indirect
 	github.com/hashicorp/errwrap v1.1.0 // indirect
+	github.com/hashicorp/go-multierror v1.1.1 // indirect
```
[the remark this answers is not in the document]
SoulPancake: Yes it is still indirect as it is using in the transformer in language/pkg/go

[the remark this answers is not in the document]
SoulPancake: This is because of pkg/go/transformer
We will maybe drop that and update it later to remove the indirect dep

SoulPancake:
We can maybe get [this](https://github.com/openfga/language/issues/600) done before we update and  merge this
so we can also remove the indirect deps

rhamzeh (commented):
I would rather not maintain our own CSV parsing implementation - a battle tested library for this case wins.

We can always move off it if needed, but for this niche case is an exception where I'm willing to tolerate a third-party lib.

At the very least if we can move that off to its own PRs so we can easily revert it.

For the multi error change, let's bump up language after it is merged in the same PR here before merging this.

senojj:
Using reflection to identify struct tags is slow and generally unnecessary for CSV; CSV formatted data records are all of the same shape for a given set, with an optional field name header. Unlike JSON which encodes fields names and other properties of the fields. For CSV encoding, I would advise using an interface along with the standard library instead. Something like:
```go
type Marshaler interface {
    MarshalCSV() ([]string, error)
}
```
And define your struct like:
```go
type Person struct {
    FirstName string
    LastName  string
    Age       int
}

func (p *Person) MarshalCSV() ([]string, error) {
    var output [3]string
    output[0] = p.FirstName
    output[1] = p.LastName
    output[2] = strconv.Itoa(p.Age)
    return output[:], nil
}
```
Then you can have a helper:
```go
func MarshalCSV(records []Marshaler, w io.Writer, header ...string) error {
    ow := csv.NewWriter(w)
    if len(header) > 0 {
        err := ow.Write(header)
        if err != nil {
            return err
        }
    }

    for _, r := range records {
        output, err := r.MarshalCSV()
        if err != nil {
            return err
        }
        err = ow.Write(output)
        if err != nil {
            return err
        }
    }
    return nil
}
```

internal/authorizationmodel/model.go
```diff
@@ -207,16 +207,16 @@ func (model *AuthzModel) ReadModelFromModFGA(modFile string) error {
 	}
 
 	moduleFiles := []language.ModuleFile{}
-	fileReadErrors := multierror.Error{}
+	fileReadErrors := []error{}
```
senojj: This incurs an allocation even when no read errors are encountered. Start with a `nil` slice, and append will allocate when needed.

SoulPancake:
Thanks alot @rhamzeh and @senojj for the feedback.
I’ll keep this PR focused on the CSV parsing changes and address the related review comments already here. I’ll move the go-multierror and language pkg bump into a separate PR.

SoulPancake:
I have addressed the comments but I'm honestly okay with keeping the third party library as well! 
However this PR might just help undo it if we ever choose to do it

internal/output/marshal.go
```diff
@@ -84,6 +86,53 @@ func (prt *csvPrinter) DisplayNoColor(data any) error {
...
+}
+
+// CSVRecordSet is implemented by a collection that can render itself as CSV:
+// a header row followed by one record per element.
+type CSVRecordSet interface {
+	CSVHeaders() []string
+	CSVRecords() []CSVMarshaler
+}
+
+func marshalCSV(data any) ([]byte, error) {
```
senojj: why do you need to accept `any` as the argument type? you seem to require a `CSVRecordSet` in the function body.

senojj: shift the type assertion to where it is necessary, in `DisplayNoColor`

cmd/tuple/read.go
```diff
@@ -48,18 +48,57 @@ type readResponse struct {
...
+}
+
+func (l readResponseCSVDTOList) CSVRecords() []output.CSVMarshaler {
+	records := make([]output.CSVMarshaler, len(l))
+	for i, dto := range l {
+		records[i] = dto
+	}
+
+	return records
+}
```
senojj: This function adds overhead that is entirely unnecessary.

senojj:
My original suggestion put the marshaling interface at the individual object instance level. That allowed for better composition than applying the serialization directly to a container type. Additionally, the original suggestion remained closer to the implementations found within the Go standard library, and thus would be more familiar to Go developers.

I'm not asking for the originally suggested code to be used verbatim, but the spirit of the approach should be maintained.

cmd/tuple/read.go
```diff
@@ -196,19 +221,22 @@ var readCmd = &cobra.Command{
...
 		if outputFormat == "csv" {
-			data, _ := response.toCsvDTO()
-
-			err := dataPrinter.Display(data)
+			records, err := response.toCsvDTO()
 			if err != nil {
+				return fmt.Errorf("failed to convert response to csv: %w", err)
+			}
+
+			if err := output.MarshalCSV(records, os.Stdout, readResponseCSVHeaders...); err != nil {
```
senojj: `MarshalCSV` calls `Write` on the provided `io.Writer` multiple times -- as it should. Each call to `Write` on `os.Stdout` is a direct system call, which will be very inefficient. You should initialize a buffered writer, pass the buffer into `MarshalCSV` and write the buffer contents to stdout as a single system call.

SoulPancake: Done, Thanks

cmd/tuple/read.go
```diff
@@ -196,19 +222,27 @@ var readCmd = &cobra.Command{
...
 		if outputFormat == "csv" {
-			data, _ := response.toCsvDTO()
-
-			err := dataPrinter.Display(data)
+			records, err := response.toCsvDTO()
 			if err != nil {
+				return fmt.Errorf("failed to convert response to csv: %w", err)
+			}
+
+			var buf bytes.Buffer
```
senojj: My apologies for not being more clear with my meaning. I had said, "write to `os.Stdout` using a single system call" and that is what you are doing here. However, for large outputs, accumulating the entire stream in memory could be unwise. Instead of using a `bytes.Buffer` use a `bufio.NewWriter(os.Stdout)` so that the writes are buffered and chunked to `os.Stdout`. Additionally, `csv.NewWriter` can take advantage of a `bufio.Writer` instance, and won't create a new one on top of it.

SoulPancake: Oops got it, no worries
used bufio.writer and removed the double wrapping

senojj (approved):

PR #713, opened 2026-06-19
fix: attach SLSA provenance to draft release via gh release upload

SoulPancake:
The slsa-github-generator (pinned at v2.1.0) attaches provenance using softprops/action-gh-release@v2.2.1, whose draft lookup reassigns the match on every page of releases instead of breaking on first hit. Once a repo crosses 100 releases (2 pages), the fresh draft on page 1 is clobbered by the empty page 2, the action falls through to createRelease(), and two drafts end up on the same tag.

Set upload-assets: false so the generator no longer touches the release, and add a release-provenance job that downloads the provenance artifact and attaches it to goreleaser's single draft via 'gh release upload', which resolves the draft deterministically by tag.

The issue was first found out in the openfga/openfga repo while attempting the same pipeline for immutable tagging 

https://github.com/openfga/openfga/pull/3178#discussion_r3443529759

## Review Checklist

## Summary by CodeRabbit

* **Chores**
  * Updated internal release workflow to optimize how provenance artifacts are handled during the draft release process. This change reorganizes artifact attachment timing and dependencies while maintaining existing verification workflows.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -237,13 +237,11 @@ jobs:
     with:
       base64-subjects-as-file: "${{ needs.goreleaser.outputs.subjects-as-file }}"
       provenance-name: "fga.intoto.jsonl"
-      # Do NOT let the generator attach the provenance to the release. It would call
-      # softprops/action-gh-release@v2.2.1, whose draft lookup reassigns the match on every
-      # page of releases instead of breaking on first hit. Once this repo has >100 releases
-      # (2 pages), the fresh draft on page 1 is clobbered by the empty page 2, the action falls
-      # through to createRelease(), and we end up with two drafts on the same tag. The provenance
-      # is always uploaded as a workflow artifact regardless, so we attach it ourselves below via
-      # `gh release upload`, which resolves the draft deterministically by tag.
+      # Don't let the generator attach the provenance:
+      # - it uses softprops/action-gh-release@v2.2.1, whose paginated draft lookup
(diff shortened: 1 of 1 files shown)


rhamzeh (dismissed):

.github/workflows/main.yaml
```diff
@@ -230,16 +230,39 @@ jobs:
...
-      upload-tag-name: ${{ github.ref_name }}
-      draft-release: true
+      # Don't let the generator attach the provenance:
+      # - it uses softprops/action-gh-release@v2.2.1, whose paginated draft lookup
+      #   creates a duplicate draft once a repo has >100 releases (2 pages)
+      # - the provenance is still published as a workflow artifact, so the
+      #   release-provenance job attaches it via `gh release upload` (by tag)
+      upload-assets: false
+
+  release-provenance:
```
curfew-marathon, line 248: Can we add a timeout-minutes on this?

SoulPancake: Yes!
Added a 10 min, should be a long enough window for download/release.
We can also re-run it if somehow it fails an attempt from the same step.

rhamzeh (dismissed):

SoulPancake:
image
Turns out it needs the contents: write access even when the upload assets flag is set to false 🤔

SoulPancake:
Known issue with it
https://github.com/slsa-framework/slsa-github-generator/issues/2044#issuecomment-1519289250

rhamzeh (approved):

PR #634, opened 2026-02-19
fix: dockers v2 for goreleaser

SoulPancake:
## Review Checklist
- [ ] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [ ] The correct base branch is being used, if not `main`
- [ ] I have added tests to validate that the change in functionality is working as expected

PR #632, opened 2026-02-18
fix: goreleaser distroless failing ( workaround )

SoulPancake:
## Description

[stackoverflow thread](https://stackoverflow.com/questions/75521775/buildx-docker-image-claims-to-be-a-manifest-list)

## Review Checklist
- [ ] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [ ] The correct base branch is being used, if not `main`
- [ ] I have added tests to validate that the change in functionality is working as expected

## Summary by CodeRabbit

* **Chores**
  * Updated Docker build configuration for amd64 and arm64 architectures.

PR #716, opened 2026-06-28
fix: publish distroless Docker image with dockers v2

### LINKED ISSUE #639: Move Docker image to distroless while preserving GoReleaser compatibility
### Problem

Currently, the Docker image is built `FROM scratch`, which omits CA certificates and causes all HTTPS (TLS) requests to fail with the following error:

```
tls: failed to verify certificate: x509: certificate signed by unknown authority
```

This issue occurs even for well-known public endpoints (e.g., https://example.com), making the CLI unusable in Docker for any TLS-enabled OpenFGA deployment.

**Steps to Reproduce**
1. Build the CLI binary:
   ```sh
   GOOS=linux GOARCH=amd64  BUILD_DIR=$PWD make build
   ```
2. Build the Docker image:
   ```sh
   docker build --platform=linux/amd64 -t fgacli -f .goreleaser.Dockerfile .
   ```
3. Run with Docker:
   ```sh
   docker run --platform=linux/amd64 -it --rm -v $PWD:/workdir fgacli --debug model write --file /workdir/model.fga --store-id=some-store-id --api-url=https://example.com
   ```
4. Observe the error:
   ```
   tls: failed to verify certificate: x509: certificate signed by unknown authority
   ```

### Solution Attempted
We tried switching to `gcr.io/distroless/static:nonroot`:

```Dockerfile
FROM gcr.io/distroless/static:nonroot
COPY fga /fga
ENTRYPOINT ["/fga"]
```

**Benefits of distroless:**
- Includes CA certificates → enables HTTPS out of the box
- Minimal attack surface, non-root default, small image size
- Recommended for static Go binaries (per Google)

### Complication: GoReleaser Incompatibility

However, when this [change](https://github.com/openfga/cli/pull/617) was applied, it caused issues during the GoReleaser publishing step:

- GoReleaser failed to create/push Docker manifests, reporting: `is a manifest list` error
- Workarounds exist (e.g., [disabling provenance](https://github.com/openfga/cli/pull/632)), but are not desirable 
- As a result, the change was reverted to restore release stability

### Request

We would like to migrate to a distroless image **while preserving GoReleaser release/publishing compatibility**. Please investigate a solution or guidance that achieves both goals:

- Enable HTTPS/TLS support for the Docker image (CA certs included)
- Ensure multi-platform release via GoReleaser continues to work without errors
- No workaround that disables provenance or key release features

#### Additional Context
Failing goreleaser logs : https://github.com/openfga/cli/actions/runs/22150410982/job/64040133213

- See earlier PR/discussion for attempted distroless migration and reversion rationale
- Reference issue relates to both release process robustness and user security/trust using the CLI via Docker

Siddhant-K-code:
## Summary
- Move the Docker image from `scratch` to `gcr.io/distroless/static-debian13:nonroot` so public CA roots are available in the container.
- Replace per-arch `dockers` + `docker_manifests` with GoReleaser `dockers_v2`, which is no longer experimental as of GoReleaser v2.16.
- Sign the published Docker image by immutable digest and select the released multi-platform digest by exact Git tag and platform metadata.

## Docs
- README Docker section now notes public CA support and calls out that private/internal CAs still need to be provided by the user.

## Validation
Ran:
- `make build`
- `go test ./...`
- `goreleaser release --clean --config .goreleaser.yaml --snapshot --skip sign,publish,announce,docker`
- focused `dockers_v2` GoReleaser validation with a local fake Docker shim
- GoReleaser v2.16 Docker signing tests for `DockerImageV2` artifact selection
- workflow digest `jq` check against a representative multi-platform `Docker Image` artifact

Reviewer command with Docker available:
- `goreleaser release --clean --config .goreleaser.yaml --snapshot --skip sign,publish,announce`

Reference: https://goreleaser.com/blog/goreleaser-v2.16/#docker-v2-is-no-longer-experimental

Fixes https://github.com/openfga/cli/issues/639

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -220,8 +220,16 @@ jobs:
             ARTIFACTS: "${{ steps.goreleaser.outputs.artifacts }}"
         run: |
           set -euo pipefail
-          artifact=$(echo "$ARTIFACTS" | jq -r '[.[] | select (.type=="Docker Manifest")][0]')
-          image=$(echo "$artifact" | jq -r '.path' | cut -d':' -f1)
+          artifact=$(
+            echo "$ARTIFACTS" | jq -er --arg image_tag "openfga/cli:${GITHUB_REF_NAME}" '
+              [
+                .[]
+                | select(.type == "Docker Image")
+                | select(.path == $image_tag)
+                | select(((.extra.Platforms // []) | index("linux/amd64")) and ((.extra.Platforms // []) | index("linux/arm64")))
+              ][0]
+            '
+          )
           digest=$(echo "$artifact" | jq -r '.extra.Digest')
           echo "digest=$digest" >> "$GITHUB_OUTPUT"
 
--- a/.goreleaser.Dockerfile
+++ b/.goreleaser.Dockerfile
@@ -1,3 +1,5 @@
-FROM scratch
-COPY fga /fga
+FROM gcr.io/distroless/static-debian13:nonroot
+
+ARG TARGETPLATFORM
+COPY ${TARGETPLATFORM}/fga /fga
 ENTRYPOINT ["/fga"]
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -73,11 +73,11 @@ docker_signs:
   - cmd: cosign
     env:
     - COSIGN_EXPERIMENTAL=1
-    artifacts: manifests
+    artifacts: images
     output: true
     args:
     - 'sign'
-    - '${artifact}'
+    - '${artifact}@${digest}'
     - "--yes" # needed on cosign 2.0.0+
 
 brews:
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -24,68 +24,25 @@ builds:
       - "-X github.com/openfga/cli/internal/build.Commit={{.Commit}}"
       - "-X github.com/openfga/cli/internal/build.Date={{.Date}}"
 
-dockers:
-  - goos: linux
-    goarch: amd64
-
+dockers_v2:
+  - images:
+      - openfga/cli
+    tags:
+      - latest
+      - "{{ .Tag }}"
+      - "v{{ .Major }}"
+      - "v{{ .Major }}.{{ .Minor }}"
     dockerfile: .goreleaser.Dockerfile
-
-    image_templates:
-      - "openfga/cli:latest-amd64"
-      - "openfga/cli:{{ .Tag }}-amd64"
-      - "openfga/cli:v{{ .Version }}-amd64"
-      - "openfga/cli:v{{ .Major }}-amd64"
-      - "openfga/cli:v{{ .Major }}.{{ .Minor }}-amd64"
-      - "openfga/cli:v{{ .Major }}.{{ .Minor }}.{{ .Patch }}-amd64"
-
-    use: buildx
-    build_flag_templates:
-      - "--platform=linux/amd64"
+    platforms:
+      - linux/amd64
+      - linux/arm64
+    labels:
+      org.opencontainers.image.created: "{{ .Date }}"
+      org.opencontainers.image.title: "{{ .ProjectName }}"
+      org.opencontainers.image.revision: "{{ .FullCommit }}"
+      org.opencontainers.image.version: "{{ .Version }}"
+    flags:
       - "--pull"
-      - "--label=org.opencontainers.image.created={{.Date}}"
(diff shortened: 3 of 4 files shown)


SoulPancake (dismissed):
dockers_v2 is currently still experimental.
There will be a stable release with goreleaser v3 launch, so we probably shouldn't use it right now

SoulPancake:
https://github.com/openfga/cli/pull/634/changes#r2828917491
https://goreleaser.com/resources/deprecations/#dockers_v2retry
FWIW it might evolve a bit more

Siddhant-K-code:
Good call. I pushed `bab9eeb` and removed `dockers_v2` completely.

This now stays on the documented `dockers` + `docker_manifests` flow and only changes the per-arch image builds from `use: buildx` to `use: docker`: https://goreleaser.com/customization/package/docker/

That keeps the existing manifest publishing shape, avoids the experimental `dockers_v2` path, and does not add the PR #632 `buildx --provenance=false` workaround. The repo's SLSA image-provenance job is unchanged.

I also folded in the smaller review notes:
- clarified that the bundled CA roots cover publicly trusted CAs, not private/internal PKI
- removed duplicate exact-version Docker tag templates (`{{ .Tag }}` is the canonical `vX.Y.Z` tag)
- removed the unused workflow variable

One caveat: GoReleaser v2.16 now marks `dockers`/`docker_manifests` as deprecated in favor of `dockers_v2`: https://goreleaser.com/deprecations#dockers. I think that is the right tradeoff here given the concern about `dockers_v2` still evolving before v3.

SoulPancake:
Oh wait, I think my earlier concern here is outdated.

This note from May means `dockers_v2` is no longer experimental: 
https://github.com/goreleaser/goreleaser/commit/e8d568690722707bcb840b20e0c972b8f00e3025
https://goreleaser.com/blog/goreleaser-v2.16/?utm_source=chatgpt.com#docker-v2-is-no-longer-experimental

I think it should be fine to move to dockers_v2 after all.Sorry for the back and forth on this, Let's do `dockers_v2`
@Siddhant-K-code

Siddhant-K-code:
Yep, that changes the decision. I pushed `c7f6dc4` and moved this back to `dockers_v2`.

I kept the other review fixes in place:
- no duplicate exact-version tags (`{{ .Tag }}` is the exact `vX.Y.Z` tag)
- README now says public CAs, with private/internal CAs called out separately
- workflow digest lookup selects the `Docker Image` artifact for the exact tag and checks both `linux/amd64` and `linux/arm64`

I also updated the PR description and added the v2.16 reference. Local focused `dockers_v2` validation is green, and the generated buildx command does not include the old `--provenance=false` workaround.

SoulPancake (dismissed):
dockers_v2 produces multiplatform image artifacts
docker_signs still targets the `artifacts: manifests`
So will this work with dockers_v2? I think it might remain unsigned

Maybe we should do something like this https://github.com/enix/x509-certificate-exporter/blob/a1a454659e5f8b403f7eeb33ad97f167a9e22e4d/.goreleaser.yaml#L304

Siddhant-K-code:
Good catch. I pushed `2796ba5` to make this explicit.

The config now uses `docker_signs.artifacts: images` and signs `${artifact}@${digest}`. That lines up with GoReleaser v2.16’s `dockers_v2` path:

- `artifacts: images` includes `artifact.DockerImageV2`: https://github.com/goreleaser/goreleaser/blob/v2.16.0/internal/pipe/sign/sign_docker.go#L56-L67
- `dockers_v2` writes image artifacts as `DockerImageV2` and stores the digest in `ExtraDigest`: https://github.com/goreleaser/goreleaser/blob/v2.16.0/internal/pipe/docker/v2/docker.go#L228-L237
- `${digest}` is populated from that artifact digest before invoking cosign: https://github.com/goreleaser/goreleaser/blob/v2.16.0/internal/pipe/sign/sign.go#L180-L185
- GoReleaser’s own Docker signing tests cover the v2 `images` case: https://github.com/goreleaser/goreleaser/blob/v2.16.0/internal/pipe/sign/sign_docker_test.go#L188-L210

Validated locally with:
- `make build`
- `go test ./...`
- `goreleaser release --clean --config .goreleaser.yaml --snapshot --skip sign,publish,announce,docker`
- focused fake-Docker `dockers_v2` run
- GoReleaser v2.16 `go test ./internal/pipe/sign -run 'TestDockerSign(Default|Artifacts)$'`

So the published image will be signed by digest during the release publish flow.

SoulPancake (approved):
LGTM
Thanks a lot @Siddhant-K-code

PR #737, opened 2026-08-05
fix!: contain store YAML file references to the store file's directory

### LINKED ISSUE #736: Store YAML nested file references are not contained to the store file's directory
### Summary

Nested file references in store YAML are resolved with `path.Join(basePath, value)` and no containment check, so a `../` prefix escapes the directory holding the store file. `fga store import` and `fga model test` then read that file.

Reported by @bugbunny-research.

### Affected code

Four fields in `internal/storetest/storedata.go`, all resolved the same way:

| field | line |
|---|---|
| `model_file` | 105 |
| `tuple_file` | 205 |
| `tuple_files[]` | 223 |
| per-test `tuple_file` | 247 |

Each joined path goes straight to `authorizationmodel.ReadFromFile` (`internal/authorizationmodel/read-from-input.go:37`) or `tuplefile.ReadTupleFile` (`internal/tuplefile/read.go:19`), both calling `os.ReadFile`.

`path.Join` runs `path.Clean`, which collapses `..`, so `path.Join("/tmp/base", "../outside/secret.txt")` yields `/tmp/outside/secret.txt`. Nothing checks the result is still under `basePath`. `StoreData.Validate()` (`storedata.go:166`) only checks user/object conflicts in test assertions. The `!filepath.IsAbs(fileName)` check at `read-from-input.go:38` guards the top-level store file path, not nested references.

Note absolute paths are already harmless: `path.Join("/base", "/etc/passwd")` returns `/base/etc/passwd`. Only `..` escapes.

### Reproduction

```bash
mkdir -p /tmp/traversal-demo/sub
echo "contents outside the store dir" > /tmp/traversal-demo/secret.txt

cat > /tmp/traversal-demo/sub/store.fga.yaml <<'YAML'
name: traversal-demo
model_file: ../secret.txt
tuples: []
tests: []
YAML

fga store import --api-url http://localhost:8080 \
  --file /tmp/traversal-demo/sub/store.fga.yaml
```

Reads `secret.txt` from one directory up and submits it as an authorization model; the API rejects it, but the file was read. Same via `tuple_file`, `tuple_files`, and a per-test `tuple_file`.

### Impact

- Reads a file outside the base directory, limited to files the invoking user can already read. Content doesn't leave the process except as a rejected API payload.
- `/dev/zero` as a file field makes `os.ReadFile` grow an unbounded buffer until OOM. A FIFO with no writer blocks instead.
- No integrity impact. Writing supplied models and tuples is what `store import` does by design. For a traversed file to be written as tuples it needs a `.json`/`.yaml`/`.yml`/`.jsonl`/`.csv` extension (`read.go:26` switches on `path.Ext`; anything else errors) and contents that parse as valid tuples.

### Fix

```go
func safeJoin(basePath, relPath string) (string, error) {
	joined := filepath.Join(basePath, relPath)

	rel, err := filepath.Rel(basePath, joined)
	if err != nil || rel == ".." || strings.HasPrefix(rel, ".."+string(filepath.Separator)) {
		return "", fmt.Errorf("file reference %q escapes the store file's directory", relPath)
	}

	return joined, nil
}
```

`filepath.Rel` is lexical, so this does not cover symlinks — that would need `filepath.EvalSymlinks` and would break intentionally symlinked fixtures, so it's a separate call.

### Tasks

- [ ] `safeJoin` helper in `internal/storetest`
- [ ] Use it at all four sites
- [ ] Tests: traversal rejected at each site, relative paths and subdirectories still work
- [ ] Decide separately on a read size limit (`/dev/zero`) and symlink containment

SoulPancake:
Closes #736

Problem
The four nested file fields in store YAML — model_file, tuple_file, tuple_files[] and per-test tuple_file — were resolved with path.Join(basePath, value) and no containment check. path.Join runs path.Clean, which collapses .., so path.Join("/tmp/base", "../secret.txt") yields /tmp/secret.txt and fga store import / fga model test would read it. StoreData.Validate() never inspected these fields.

There was also no guard on the type of file a reference pointed at, which is the availability half of the report. A FIFO with no writer blocks os.ReadFile forever, hanging the process. An endless device such as /dev/zero reads a large amount before failing — worth noting the report described this as an unbounded read leading to OOM, but in practice it terminates once YAML parsing hits a control character, so it wastes time rather than exhausting memory. Both are reachable through any of the four fields.

Fix
References are read through an os.Root handle on the base directory instead of being resolved with path.Join, which rejects .. traversal, absolute paths, and symlinks pointing outside the tree.

The read goes through the same handle that validated the reference, rather than validating a reference and then resolving it to a path for someone else to open. That distinction matters: a lexical path and a root-relative lookup do not always agree. Given a symlink base/link -> sub/dir, os.Root resolves link/../target.json to base/sub/target.json while filepath.Join collapses it to base/target.json. Making the latter a symlink out of the tree means a check against the former passes while the read escapes — so validating one path and reading another leaves the containment bypassable. safefile.ReadContained therefore opens and reads through the handle, and re-stats the returned descriptor rather than the name, so the file that is read is the file that was checked. A filepath.Rel check, being purely lexical, would not have covered this at all.

A referenced target must also be a regular file. A metadata-only Stat runs first, since opening a FIFO would block before any check could reject it, and the opened descriptor is re-checked before its contents are read. Between containment and the file-type check, both DoS shapes are refusedrefused for nested references: an in-tree FIFO is rejected on mode, and /dev/zero (or a symlink to it) is rejected as an escape.

The file named on the command line is deliberately not subject to the file-type check. The reported issue is about the files a store YAML references, not the file the user passed, and checking it broke reading a store file from a pipe — fga model test --tests <(...) and fga store import --file <(...) both work on main.

A modular model referenced from a store file reads its fga.mod contents entries at parse time, separately from the store file's own references, so those reads are contained too — LoadModel records the base and it travels with the store data to be enforced there. (A literal .. entry is already rejected by language.TransformModFile; the case this closes is a plain-named entry whose file is a symlink out of the tree.) Naming an fga.mod directly on the command line is unaffected. Because fga.mod contents entries are slash-separated while modFile is an OS path, entries are joined with filepath.Dir/filepath.Join after filepath.FromSlash, so a backslash-separated .fga.mod path on Windows resolves its module files next to the mod file rather than against the working directory.

Errors are joined per-file, so one bad reference reports and the rest still load.

Scope: nested references only
Containment and the regular-file check apply to the four nested fields, not to the store file the user names on the command line. That file is still opened with os.Open and no file-type check, so fga model test --tests /dev/zero or a FIFO passed as --tests is not refused by this PR.

That is deliberate, and it is the reason the file-type check is not applied there: the reported issue is about the files a store YAML references, not the file the user passed, and checking the top-level file broke reading a store file from a pipe — fga model test --tests <(...) and fga store import --file <(...) both work on main and still work here. Anyone who can pass --tests already chooses the process's arguments, so a self-inflicted hang is a different threat model from a store file that reads paths its author never intended.

Breaking change
References that resolve outside the store file's directory are now rejected by default. A store file like this stops working:

# tests/fixtures/relative-path/relative-path-store.fga.yaml
name: Relative Path Store
model_file: ../basic-model.fga
tuple_file: ../basic-tuples.json
This pattern is legitimate and in use — sharing one model or tuple file across several store files in sibling directories. Both commands take --allow-external-files to opt back in:

fga store import --file ./store.fga.yaml --allow-external-files
fga model test --tests ./store.fga.yaml --allow-external-files
With the flag set, containment is skipped but the regular-file check still applies.

With the flag set, containment is skipped but the regular-file check still applies. An absolute reference is then used as-is rather than joined onto the store file's directory, so model_file: /shared/models/basic-model.fga resolves to that path.

The commit carries a BREAKING CHANGE: footer, so this releases as a minor bump.

Tests
internal/storetest/security_test.go — containment: reference inside base resolves, .. traversal blocked, absolute path blocked, traversal permitted when external files are allowed, an absolute reference resolved as-is when external files are allowed, and the symlink/.. divergence above cannot escape.
internal/storetest/security_unix_test.go — a real mkfifo FIFO is rejected in both modes. Unix-only build tag, since syscall.Mkfifo does not exist on Windows and a runtime GOOS skip still breaks the build there.
tests/model-test-cases.yaml / tests/import-tests-cases.yaml — the existing relative-path cases now assert failure by default and success with --allow-external-files; a new tests/fixtures/traversal/traversal-store.fga.yaml covers an escape attempt.
go build ./... and go test ./... pass.

Performance
Reads pre-size their buffer from the Stat that already ran, rather than using io.ReadAll, which starts small and repeatedly doubles. Allocation is at parity with os.ReadFile (5.25MB for a 5MB file; io.ReadAll used 11.0MB). The remaining cost is the extra OpenRoot/Stat syscalls, roughly 120µs per referenced file, once per file at load time.

fga tuple write and fga tuple delete read tuple files through the original path and are unchanged.

Notes
Absolute paths were already harmless with path.Join — path.Join("/base", "/etc/passwd") returns /base/etc/passwd, since an absolute second argument is appended rather than honoured. os.Root rejects them explicitly now regardless.
os.Root requires Go 1.24; go.mod is on 1.25.7.
Summary by CodeRabbit
New Features

Added --allow-external-files to store import and model test.
External referenced files remain disabled by default and require explicit opt-in.
Bug Fixes

Blocked path traversal, external references, unsafe symlinks, and non-regular files during file loading.
Blocked path traversal, external references, unsafe symlinks, and non-regular files for files referenced from within a store file.
Added clear errors when referenced files are inaccessible or unsafe.
Documentation

Documented the new option, default behavior, and associated trust warning.


internal/storetest/storedata.go
```diff
@@ -88,7 +90,59 @@ type StoreData struct {
...
+	info, err := root.Stat(ref)
+	if err != nil {
+		return "", fmt.Errorf("file reference %q is not accessible within %q: %w", ref, basePath, err)
+	}
+
+	if !info.Mode().IsRegular() {
+		return "", fmt.Errorf("file reference %q (mode %s): %w", ref, info.Mode(), safefile.ErrNotRegularFile)
+	}
+
+	return filepath.Join(basePath, ref), nil
```
Siddhant-K-code: `root.Stat` and `filepath.Join` do not resolve `symlink/..` the same way. With `link -> sub/dir` and `link/../target.json`, the stat checks `sub/target.json` but `filepath.Join` returns `target.json`, which can be an escaping symlink. 

Keep the `Root`/opened file through the read instead of returning a normal path.

internal/storetest/storedata.go
```diff
@@ -98,11 +152,16 @@ func (storeData *StoreData) LoadModel(basePath string) (authorizationmodel.Model
...
+	if err != nil {
+		return format, err
+	}
+
 	var inputModel string
 
 	storeName := storeData.Name
 	if err := authorizationmodel.ReadFromFile(
-		path.Join(basePath, storeData.ModelFile),
+		modelPath,
```
Siddhant-K-code, line 177: This validates only the `.fga.mod` file. Later `ReadModelFromModFGA` reads each `contents` entry with `path.Join` and `os.ReadFile`, so a `../outside.fga` entry bypasses `--allow-external-files=false`. 

Resolve module entries through the same containment mechanism.

internal/storetest/security_test.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	t.Parallel()
+
+	base := t.TempDir()
+	fifo := filepath.Join(base, "pipe")
+
+	if err := syscall.Mkfifo(fifo, 0o600); err != nil {
+		t.Skipf("unable to create FIFO: %v", err)
+	}
```
Siddhant-K-code: `syscall.Mkfifo` is not defined on Windows, so the runtime skip cannot prevent this test from failing to compile there. 

Put the FIFO test/helper behind a Unix build tag.

Siddhant-K-code (dismissed):
Please address the three inline findings before merge:

- `os.Root` validation is discarded before the actual read, allowing a symlink/`..` resolution mismatch.
- Nested files referenced by `.fga.mod` bypass containment.
- The FIFO test does not compile on Windows.

SoulPancake:
All three fixed, thanks.

The `root.Stat` / `filepath.Join` divergence was the real one — I reproduced the full exfiltration with `base/link -> sub/dir` and `base/target.json -> ../outside-secret.json`: the stat resolves to `base/sub/target.json` and passes while the read followed `base/target.json` out of the tree. `resolveFile` no longer returns a path; reads go through the same `os.Root` handle that validated them, re-statting the descriptor rather than the name. Regression test added.

On `fga.mod`, the `..` vector doesn't reproduce language basically `language.TransformModFile` rejects it with `invalid contents item ../outside.fga` before any read. What did get through is a plain-named contents entry whose file is a symlink out of the tree, so I've contained those reads too: `LoadModel` records the base (modular models defer reading module files until parse time) and each entry resolves through `os.Root`. Naming an `fga.mod` directly is unchanged.

FIFO test moved to `security_unix_test.go` behind `//go:build unix` — `GOOS=windows go vet` failed before, passes now.

@Siddhant-K-code

Siddhant-K-code:
@SoulPancake - Lints are failing now, can you please fix that too?

SoulPancake:
Quickly evaluating if there is a performance/allocation regression with these

internal/storetest/read-from-input.go
```diff
@@ -50,13 +49,6 @@ func ReadFromFile(
 		absFileName = filepath.Join(basePath, fileName)
 	}
 
-	// Reject non-regular files before reading: a FIFO with no writer blocks the
-	// read forever, and an endless device such as /dev/zero grows the read buffer
-	// until the process is OOM-killed.
-	if err := safefile.CheckRegular(absFileName); err != nil {
-		return format, nil, fmt.Errorf("cannot read store file: %w", err)
-	}
```
Siddhant-K-code: This restores process substitution, but it also removes the top-level special-file protection. `--file` / `--tests` can again point to a FIFO or `/dev/zero`, causing a block or unbounded read.

If top-level inputs are intentionally exempt, update the PR description and docs. Otherwise, keep the protection and provide an explicit opt-in for process substitution.

internal/safefile/safefile.go
```diff
@@ -0,0 +1,162 @@
...
+	if err := checkRegularInfo(name, info); err != nil {
+		return nil, err
+	}
+
+	// Pre-size from the descriptor's reported size and grow only if the file
+	// turned out to be larger, which is what os.ReadFile does. io.ReadAll would
+	// instead start small and repeatedly double, allocating roughly twice the
+	// file size in the process.
+	data := make([]byte, 0, info.Size()+1)
+
```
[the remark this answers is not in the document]
Siddhant-K-code: `make([]byte, 0, info.Size()+1)` accepts `int64`; `make` accepts any integer-typed size argument. [Go specification](https://go.dev/ref/spec#Making_slices_maps_and_channels)

internal/storetest/read-from-input.go
SoulPancake: Give me a bit, working on trying some options

SoulPancake: Updated desc
Don't think we should do it

internal/authorizationmodel/model.go
```diff
@@ -376,6 +359,70 @@ func (model *AuthzModel) DisplayAsDSL(fields []string) (*string, error) {
...
+	parsedModFile, err := language.TransformModFile(string(modFileContents))
+	if err != nil {
+		return fmt.Errorf("failed to transform fga.mod file due to %w", err)
+	}
+
+	moduleFiles := []language.ModuleFile{}
+
+	var fileReadErrors []error
+
+	directory := path.Dir(modFile)
```
Siddhant-K-code: `modFile` is an OS filesystem path, but `path.Dir` and `path.Join` only handle `/`. 

On Windows, a backslash-separated `.fga.mod` path resolves its module files from the working directory.

Use `filepath.Dir`, `filepath.Join`, and `filepath.FromSlash` for manifest entries.

internal/storetest/storedata.go
```diff
@@ -86,10 +87,70 @@ type StoreData struct {
...
+		return joined, data, nil
+	}
+
+	data, err := safefile.ReadContained(basePath, ref)
+	if err != nil {
+		return "", nil, fmt.Errorf("file reference %w", err)
+	}
+
+	return joined, data, nil
 }
```
Siddhant-K-code, line 148: `--allow-external-files` still joins absolute references to `basePath`, so an absolute path does not resolve externally.

Use `ref` directly when `filepath.IsAbs(ref)` is true, and add coverage for that case.

Siddhant-K-code (dismissed):
Please address the two inline findings before merge:

- Modular model paths do not work correctly on Windows.
- `--allow-external-files` does not support absolute external references.

Also update the PR description: it claims top-level store files are protected from special-file reads, but the final implementation only protects nested references.

internal/authorizationmodel/model.go
SoulPancake: Fixed in 73c054a. `directory` now comes from `filepath.Dir(modFile)`, and each `contents` entry goes through `filepath.FromSlash` before `filepath.Join`, since entries inside `fga.mod` are always slash-separated while `modFile` is an OS path. The `path` import is gone from the file, so the remaining path handling here (including `filepath.Rel` in `readModelFile`) is uniformly OS-aware.

internal/storetest/storedata.go
SoulPancake: Fixed in 4c307a4. In external mode the reference is used as-is when `filepath.IsAbs(ref)`, and only joined onto `basePath` when relative, so `model_file: /shared/models/basic-model.fga` now resolves to that path instead of `<basePath>/shared/models/basic-model.fga`. The returned path (used for format detection) is the same resolved path, so extension-based detection still lines up. Contained mode is unchanged — an absolute reference is still rejected there by `os.Root`.

Covered by a new subtest in `internal/storetest/security_test.go`, `absolute path is allowed when external files are permitted`, which asserts both the resolved path and the contents; the existing `absolute path is blocked by default` case pins the contained behaviour.

cmd/store/import.go
```diff
@@ -97,7 +97,7 @@ func updateStore(
 	authModel := authorizationmodel.AuthzModel{}
 	clientConfig.StoreID = storeID
 
-	if err := authModel.ReadModelFromString(storeData.Model, format); err != nil {
+	if err := authModel.ReadModelFromStringContained(storeData.Model, format, storeData.ModelContainBase()); err != nil {
```
Siddhant-K-code, line 104: It looks like containment is only applied to the update path here. For imports without `--store-id`, `createStore` still calls `CreateStoreWithModel`, which parses the model without `ModelContainBase()`. As a result, an `fga.mod` contents entry could still follow a symlink outside the store directory.

Could we pass `ModelContainBase()` through the create path as well, while keeping direct `store create` behavior unchanged? A regression test covering import without `--store-id` would help protect both paths.

Siddhant-K-code (commented):
Overall seems fine now. Do you think, we should add a callout regarding breaking change in CHANGELOG?

SoulPancake:
@Siddhant-K-code 
Yes, We will do so.
Let me change the PR title.
We will do a minor bump too with this and edit the release-please generated changelog with a proper note about this.

Siddhant-K-code (approved):
LGTM.

Just make sure to update changelog regarding breaking change

PR #738, opened 2026-08-05
fix: remove go-safecast dependency

### LINKED ISSUE #672: fix: remove go-safecast dependency
Remove the github.com/rung/go-safecast dependency by replacing two safecast.Int32() calls with inline bounds checks using math.MaxInt32.

**File:** internal/tuple/import.go:107–115

**Changes:**
- Replace safecast.Int32(maxTuplesPerWrite) with inline bounds check
- Replace safecast.Int32(maxParallelRequests) with inline bounds check
- Remove go-safecast import

Siddhant-K-code:
## Summary

- replace `go-safecast` conversions with explicit `math.MaxInt32` bounds checks in shared import-parameter validation
- remove `go-safecast` from module and lint configuration
- add coverage for values exceeding the `int32` range, including 32-bit compilation behavior

Fixes #672

## Testing

- `go mod verify`
- `go test ./...`
- `GOARCH=386 go test ./internal/tuple -run TestImportTuplesRejectsValuesOutsideInt32Range -count=1`

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to reject import settings that exceed supported 32-bit integer limits.
  * Improved error reporting for invalid tuple import configuration values.

* **Tests**
  * Added coverage for out-of-range import settings across supported platforms.


Siddhant-K-code:
Addressed the 32-bit compilation comment in `9316192`. The test now increments a non-constant `int` value only after the 32-bit skip. Verified with `GOARCH=386 go test ./internal/tuple -run TestImportTuplesRejectsValuesOutsideInt32Range -count=1`.

ewanharris (approved):

PR #742, opened 2026-08-16
fix: skip non-regular files matched by glob in model test

### LINKED ISSUE #739: fga model test hangs when a glob matches a FIFO
### Summary

`fga model test --tests` accepts a glob (`filepath.Glob`, `cmd/model/test.go:75`) and loops over every match. If a matched file is a FIFO with no writer, `os.ReadFile` blocks waiting for data that never arrives and the CLI hangs until it is killed.

### Reproduction

```bash
mkdir repo && cd repo

cat > ok.fga.yaml <<'YAML'
name: ok
model: |
  model
    schema 1.1
  type user
tuples: []
tests: []
YAML

mkfifo evil.fga.yaml

fga model test --tests "*.fga.yaml"
```

Hangs indefinitely; needs Ctrl-C or SIGKILL. `evil.fga.yaml` was never named — the glob picked it up.

### Reachability

A FIFO can't be committed to git (`git add` skips it), so this doesn't survive `git clone`. What does work:

- an extracted tarball or zip (`tar` preserves FIFOs)
- a mounted volume or shared directory
- a committed symlink to a FIFO path that already exists on the victim's machine (symlinks do survive `git clone`)

Impact is a wedged terminal or CI job. A symlink to `/dev/zero` is also reachable but doesn't hang — the read terminates and YAML parsing fails with `control characters are not allowed`.

### Prior art

- kubectl doesn't check file type. `ExpandPathsToFileVisitors` (`cli-runtime/pkg/resource/visitor.go`) filters only directories and extensions during `filepath.Walk`, and `FileVisitor.Visit` does a bare `os.Open`. It streams rather than buffers.
- Terraform checks, and distinguishes named paths from glob matches (`internal/lang/funcs/filesystem.go`). `fileexists` errors per irregular type — "is a device node", "is a named pipe", "is a unix domain socket". `fileset`, which globs, does `os.Stat` per match and `continue`s past anything where `!fi.Mode().IsRegular()`.

### Suggested direction

Skip non-regular matches after `filepath.Glob`, following `fileset`. That leaves explicitly named paths alone, so `--tests <(...)` and `--file <(...)` keep working.

Whether an explicitly named non-regular path should error or be read as it is today is a separate decision.

terry-writer:
Closes #739

## Problem
`fga model test --tests` expands the `--tests` flag via `filepath.Glob`
and reads every match with `os.ReadFile`. If a match is a FIFO with no
writer, the read blocks forever and the CLI hangs until killed.

## Fix
After `filepath.Glob`, each match is now `os.Stat`'d and filtered to
regular files only (following the approach used by Terraform's
`fileset`, referenced in the issue). Non-regular matches (FIFOs,
devices, sockets) are silently skipped rather than read.

The literal-path fallback (used when the glob pattern matches nothing)
is untouched, so explicitly named non-regular paths — e.g. process
substitution like `--tests <(...)` — continue to work exactly as
before.

The filtering logic is extracted into `resolveTestFiles` so it can be
unit tested without mocking the full FGA client.

## Open question for maintainers
If **every** glob match is non-regular, `resolveTestFiles` now returns
an empty slice, and the caller falls through to the literal-path
fallback — which will `os.Stat` the *pattern itself* (e.g.
`*.fga.yaml`) and fail with "test file does not exist". The issue
notes that whether an explicitly named non-regular path should error
or be read "as it is today" is a separate decision; I went with the
narrower, non-breaking fix here and left that question open rather
than guessing at the intended behavior.

## Testing
- Added `cmd/model/test_test.go` with 3 unit tests covering:
  - a FIFO mixed with a regular file among glob matches (repro of the
    issue, using `syscall.Mkfifo`)
  - no glob matches (empty result, not an error)
  - multiple regular files mixed with a FIFO
- `make test-unit` and `make lint` pass locally.
- Manually reproduced the original hang with `mkfifo` + `fga model
  test --tests "*.fga.yaml"`, confirmed it no longer hangs after the
  fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved test file pattern handling by excluding non-regular files and reporting invalid patterns or file access errors more reliably.
  * Unmatched patterns now complete without errors, while literal file paths continue to work as before.

* **Tests**
  * Added coverage for unmatched patterns, multiple matching files, and special files that should be skipped.

### DIFF
--- a/cmd/model/test.go
+++ b/cmd/model/test.go
@@ -36,6 +36,9 @@ import (
 var errAllTestFilesNonRegular = errors.New("tests pattern matched only non-regular files (e.g. FIFOs); " +
 	"pass an explicit regular file path instead")
 
+// errNoTestFilesMatched is returned when a tests glob pattern matches no files at all.
+var errNoTestFilesMatched = errors.New("no test files matched pattern")
+
 // modelTestCmd represents the test command.
 var modelTestCmd = &cobra.Command{
 	Use:     "test",
@@ -166,28 +169,37 @@ var modelTestCmd = &cobra.Command{
 	},
 }
 
-// resolveTestFiles expands testsPattern via filepath.Glob and filters out any matches that
-// are not regular files (e.g. FIFOs, devices, sockets). Reading a FIFO with no writer via
-// os.ReadFile blocks forever, so non-regular glob matches are silently skipped rather than
-// read.
+// resolveTestFiles turns testsPattern into the list of test files to read.
 //
-// If the pattern matches nothing at all, it is treated as a literal path (e.g. to keep
-// process substitution like --tests <(...) working) and is not subject to the regular-file
-// check. If the pattern matches only non-regular files, that is treated as an error rather
-// than silently falling back to the literal-path behavior, since testsPattern itself
-// (e.g. "*.fga.yaml") is very unlikely to also be a valid literal path.
+// A pattern with no glob metacharacters (*, ?, [) is treated as an explicit, literal path
+// and honored as-is - including non-regular paths such as process substitution
+// (--tests <(...)), which resolves to a FIFO like /dev/fd/11. The caller asked for that
+// exact path, so it is not subject to the regular-file filter below.
+//
+// A pattern with glob metacharacters is expanded via filepath.Glob, and non-regular matches
+// (FIFOs, devices, sockets) are dropped: a glob is not an explicit request for any single
+// file, and reading a FIFO with no writer via os.ReadFile blocks forever. If the glob matches
+// only non-regular files, that is an error rather than something to read.
+//
+// The metacharacter check (rather than, say, comparing the single glob match against the
+// pattern) is deliberate: a FIFO literally named "*.fga.yaml" must still be filtered out, not
+// mistaken for a literal path.
 func resolveTestFiles(testsPattern string) ([]string, error) {
+	if !strings.ContainsAny(testsPattern, `*?[`) {
+		if _, statErr := os.Stat(testsPattern); statErr != nil {
+			return nil, fmt.Errorf("test file %s does not exist: %w", testsPattern, statErr)
+		}
+
+		return []string{testsPattern}, nil
+	}
+
 	rawMatches, err := filepath.Glob(testsPattern)
 	if err != nil {
 		return nil, fmt.Errorf("invalid tests pattern %s due to %w", testsPattern, err)
 	}
 
 	if len(rawMatches) == 0 {
-		if _, statErr := os.Stat(testsPattern); statErr != nil {
-			return nil, fmt.Errorf("test file %s does not exist: %w", testsPattern, statErr)
-		}
-
-		return []string{testsPattern}, nil
+		return nil, fmt.Errorf("%w: %s", errNoTestFilesMatched, testsPattern)
 	}
 
 	regularFileNames := rawMatches[:0]
--- a/cmd/model/test.go
+++ b/cmd/model/test.go
@@ -87,7 +87,6 @@ var modelTestCmd = &cobra.Command{
 		}
 
 		multipleFiles := len(fileNames) > 1
-
 		clientConfig := cmdutils.GetClientConfig(cmd)
 
 		fgaClient, err := clientConfig.GetFgaClient()
@@ -167,6 +166,35 @@ var modelTestCmd = &cobra.Command{
 
 		return nil
 	},
+}
+
+// resolveTestFiles expands testsPattern via filepath.Glob and filters out any matches that
+// are not regular files (e.g. FIFOs, devices, sockets). Reading a FIFO with no writer via
+// os.ReadFile blocks forever, so non-regular glob matches are silently skipped rather than
+// read. This filtering only applies to glob matches: if the pattern matches nothing, the
+// pattern itself is treated as a literal path and returned as-is, untouched by the regular
+// file check (e.g. to keep process substitution like --tests <(...) working).
+
+func resolveTestFiles(testsPattern string) ([]string, error) {
+	fileNames, err := filepath.Glob(testsPattern)
+	if err != nil {
+		return nil, fmt.Errorf("invalid tests pattern %s due to %w", testsPattern, err)
+	}
+
+	regularFileNames := fileNames[:0]
+
+	for _, name := range fileNames {
+		info, statErr := os.Stat(name)
+		if statErr != nil {
+			return nil, fmt.Errorf("failed to stat test file %s: %w", name, statErr)
+		}
+
+		if info.Mode().IsRegular() {
+			regularFileNames = append(regularFileNames, name)
+		}
+	}
+
+	return regularFileNames, nil
 }
 
 func init() {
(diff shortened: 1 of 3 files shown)


SoulPancake:
@terry-writer Can you please sign the CLA?

SoulPancake (changes requested):
Thanks for working on this.

- `resolveTestFiles` is tested, but `modelTestCmd` still calls `filepath.Glob` directly. I ran the FIFO reproduction against both `main` and this PR, and both commands were still blocked when killed after three seconds. Wiring the resolver into the command made the same reproduction complete successfully. Could you update `modelTestCmd` to use it and add a test through the actual command path? A straight replacement also double-wraps resolver errors and treats “all matches filtered” like an unmatched glob, so those cases need to be handled separately.

- `cmd/model/test_test.go` uses `syscall.Mkfifo` without a Unix build constraint. I verified that the model test package compiles for Windows on the base commit but fails on this PR with `undefined: syscall.Mkfifo` at lines 25 and 82. Could you move the FIFO-specific tests into a `*_unix_test.go` file with `//go:build unix`, following the existing pattern in `internal/storetest/security_unix_test.go`?

terry-writer:
Thanks for the thorough review!

- Wired `RunE` to actually call `resolveTestFiles` instead of calling
  `filepath.Glob` directly — the earlier fix wasn't taking effect at
  runtime, good catch. Also folded the literal-path fallback into
  `resolveTestFiles` itself, so "glob matched nothing" and "glob
  matched only non-regular files" are now distinguished, avoiding the
  double-wrapped error and the confusing fallback-to-literal-path
  message you flagged.
- Moved the FIFO-based tests into `test_unix_test.go` behind
  `//go:build unix`, following the pattern in
  `internal/storetest/security_unix_test.go`. Added a
  platform-agnostic `test_test.go` for the non-FIFO cases.

`make build`, `make test-unit`, and `make lint` all pass locally.

SoulPancake (commented):
Thanks, the glob fix works now and the Windows compile issue is sorted.

One problem left: this breaks process substitution and explicit FIFO paths.

```console
$ fga model test --tests <(cat ok.fga.yaml)
Error: tests pattern matched only non-regular files (e.g. FIFOs); pass an explicit regular file path instead: /dev/fd/11
```

The literal-path fallback never runs for paths that exist. `filepath.Glob` doesn't return zero matches for a plain path, it returns the path itself as a single match. So `/dev/fd/11` goes through the regular-file filter and gets rejected.

Suggest checking for glob metacharacters up front instead:

```go
func resolveTestFiles(testsPattern string) ([]string, error) {
	// No glob metacharacters: treat as a literal path and honor it as-is so
	// explicitly named non-regular paths (e.g. --tests <(...)) keep working.
	if !strings.ContainsAny(testsPattern, `*?[`) {
		if _, statErr := os.Stat(testsPattern); statErr != nil {
			return nil, fmt.Errorf("test file %s does not exist: %w", testsPattern, statErr)
		}

		return []string{testsPattern}, nil
	}

	rawMatches, err := filepath.Glob(testsPattern)
	if err != nil {
		return nil, fmt.Errorf("invalid tests pattern %s due to %w", testsPattern, err)
	}

	regularFileNames := rawMatches[:0]

	for _, name := range rawMatches {
		info, statErr := os.Stat(name)
		if statErr != nil {
			return nil, fmt.Errorf("failed to stat test file %s: %w", name, statErr)
		}

		if info.Mode().IsRegular() {
			regularFileNames = append(regularFileNames, name)
		}
	}

	if len(regularFileNames) == 0 {
		return nil, fmt.Errorf("%w: %s", errAllTestFilesNonRegular, testsPattern)
	}

	return regularFileNames, nil
}
```

Avoid a `rawMatches[0] == testsPattern` equality check here: a FIFO literally named `*.fga.yaml` as the only match would pass it and hang again. The metacharacter check filters that correctly. This also drops the odd "stat the glob pattern as a path" fallback, so an unmatched glob gets a sensible error instead of `stat *.fga.yaml: no such file`.

Two tests to add:
- a literal FIFO path is returned as-is (the current literal-path test uses a regular file, so it can't catch this)
- a FIFO named `*.fga.yaml` gets filtered, not treated as literal

The `resolveTestFiles` doc comment needs updating too, it credits process substitution to the no-match fallback, which isn't what happens.

Not blocking: stat-then-read still has a TOCTOU window. A follow-up could route the top-level read through `internal/safefile` like #737 did for nested refs.

terry-writer:
Thanks again! Addressed all three points:

- `RunE` now calls `resolveTestFiles` instead of `filepath.Glob`
  directly, so the fix takes effect on the real command path. Also
  folded the literal-path fallback into `resolveTestFiles` so
  "matched nothing" and "matched only non-regular files" are handled
  distinctly, avoiding the double-wrapped error.
- Added `TestModelTestCmdDoesNotHangOnFifoGlobMatch`, which drives
  `modelTestCmd.Execute()` with a FIFO glob match and fails if it
  doesn't return within a deadline — covering the actual command
  path rather than just the resolver.
- Moved the FIFO tests into `test_unix_test.go` behind
  `//go:build unix`, following `security_unix_test.go`.

`make build`, `make test-unit`, and `make lint` all pass.

terry-writer:
Good catch — you're right, the no-match fallback never fired for
existing paths, so process substitution was broken. Fixed by
branching on glob metacharacters (*, ?, [) up front: a pattern with
none is treated as an explicit literal path and honored as-is
(including FIFOs from <(...)), and only patterns with metacharacters
go through glob expansion + the regular-file filter. That keeps a
FIFO literally named "*.fga.yaml" filtered rather than read, as you
noted.

Added the two tests you suggested (explicit FIFO path honored; FIFO
named like a glob still filtered) and updated the doc comment.
Verified `--tests <(cat ok.fga.yaml)` now exits 0 instead of being
rejected.

On the TOCTOU point: agreed it's worth routing the top-level read
through internal/safefile like #737 — happy to do that as a
follow-up PR so this one stays focused. Thanks for the thorough
reviews throughout.

SoulPancake (approved):
Thanks @terry-writer 
LGTM

terry-writer:
Picking up the TOCTOU follow-up we discussed — before I open a PR I
wanted to check the direction with you, since the obvious approach
looks like it would undo a deliberate decision from #737.

Routing the top-level read through `safefile.ReadExternal` would
reintroduce the regular-file check on the file the user names on the
command line. #737 explicitly removed that ("keep process
substitution working"), because `--tests <(...)` and
`--file <(...)` resolve to a FIFO and were failing with "not a
regular file". So a straight swap would break process substitution
again — the same thing we just fixed in #742.

If I read the intent correctly, closing the TOCTOU window on the
top-level read means keeping "accept whatever path the user named,
regardless of file type" while still doing the read safely:
nonblocking open plus the descriptor re-check that `readOpened`
already does, minus the `IsRegular` gate. That would need something
like a third entry point in `safefile` (or an option on the existing
one) rather than reusing `ReadExternal`.

Does that match what you had in mind? And is it worth doing at all
given the top-level file is explicitly user-named — happy to drop it
if you'd rather not grow the safefile API for this.

PR #745, opened 2026-08-18
feat: add --name filter to store list

### LINKED ISSUE #552: Allow filtering stores by name
The OpenFGA API supports filtering stores by names. The CLI should also support it, e.g.

```
fga store list --name "Dev"
```

Note that the Go SDK does not expose that functionality yet, so it needs to be done there first.

https://github.com/openfga/go-sdk/issues/186

ChrisJr404:
Adds a `--name` flag to `fga store list` so you can filter stores by name, matching the filter the ListStores API already supports.

The flag binds to a string that gets passed through as `Name` on `ClientListStoresOptions`. It's only set when non-empty, so the default `fga store list` behaves exactly as before and the existing pagination path is untouched.

```
fga store list --name my-store
```

Fixes #552. The issue was assigned back in Hacktoberfest 2025 but no PR ever came of it, so picking it up here.

### Testing
Added `TestListStoresWithName` which asserts `--name` reaches the SDK options via the gomock `Options(...)` expectation, alongside the existing list tests (updated for the new param). `go build ./...`, `go vet ./cmd/store/...`, `gofmt -l`, and `go test ./cmd/store/...` all pass. Also updated the store-list parameter list in the README.

## Summary by CodeRabbit

* **New Features**
  * Added a `--name` filter to the `fga store list` command.
  * Users can now list stores matching a specified name.

* **Documentation**
  * Documented the new `--name` option and its usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -364,6 +364,7 @@ fga store **list**
 
 ###### Parameters
 * `--max-pages`: Max number of pages to retrieve (default: 20)
+* `--name`: Filter stores by name
 
 ###### Example
 `fga store list`
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ custom-headers:
 | [Create a Store](#create-store) | `create` | `--name`        | `fga store create --name="FGA Demo Store"`               |
 | [Import a Store](#import-store) | `import` | `--file`        | `fga store import --file store.fga.yaml`                 |
 | [Export a Store](#export-store) | `export` | `--store-id`    | `fga store export --store-id=01H0H015178Y2V4CX10C2KGHF4` |
-| [List Stores](#list-stores)     | `list`   |                 | `fga store list`                                         |
+| [List Stores](#list-stores)     | `list`   | `--name`        | `fga store list --name="FGA Demo Store"`                 |
 | [Get a Store](#get-store)       | `get`    | `--store-id`    | `fga store get --store-id=01H0H015178Y2V4CX10C2KGHF4`    |
 | [Delete a Store](#delete-store) | `delete` | `--store-id`    | `fga store delete --store-id=01H0H015178Y2V4CX10C2KGHF4` |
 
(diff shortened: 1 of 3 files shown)


SoulPancake:
👋 @ChrisJr404 Can you please sign the EasyCLA

cmd/store/list.go
```diff
@@ -89,4 +102,5 @@ var listCmd = &cobra.Command{
 
 func init() {
 	listCmd.Flags().Int("max-pages", MaxStoresPagesLength, "Max number of pages to get.")
+	listCmd.Flags().String("name", "", "Filter stores by name.")
```
Siddhant-K-code: The API performs exact-name matching substrings and regexes are not supported. Could we make that explicit here and in the README, e.g. `Filter stores by exact name.`?

README.md
```diff
@@ -364,6 +364,7 @@ fga store **list**
 
 ###### Parameters
 * `--max-pages`: Max number of pages to retrieve (default: 20)
+* `--name`: Filter stores by name
```
Siddhant-K-code: Could we also update the “List Stores” summary row at line 210 to include `--name` and a filtered example? It currently suggests this command has no parameters.

Siddhant-K-code (commented):
@ChrisJr404, overall it looks good, just some suggestion comments (above)

ChrisJr404:
Thanks, updated the help text to say it filters by exact name, and added --name plus a filtered example to both the summary row and the List Stores section in the README.

Siddhant-K-code (approved):

SoulPancake (commented):
Tiny nit, totally non-blocking: the cobra `Example:` in `list.go` still only shows `fga store list`, so folks running `--help` won't see the new flag in action (the README got the filtered example, but help text is where most people look first). Other commands like `store create` and `tuple write` already use multi-line examples, so something like:

```go
Example: `fga store list
fga store list --name "FGA Demo Store"`,
```

would give help-text parity.
 Nice work! 🎉

PR #761, opened 2026-09-16
feat: add fga mapping run command

ewanharris:
The mapping group can author, validate and test a mapping but can't show what tuples a JSON payload actually produces.

#### How is it being solved?
`fga mapping run` evaluates a mapping against JSON input and emits the resulting tuple operations, so it composes with `fga tuple write`.

- `run` subcommand added to the `fga mapping` group
- JSONL input (one object per line) from stdin or `--input`
- Default streaming JSONL with an `op` field; `--writes-only` emits the bare `ClientTupleKey` shape for `fga tuple write --file`; `--format json` emits a batch splitting writes/deletes
- `--aggregate` buffers the run and collapses it (cross-record dedup + write/delete conflict detection)
- `--continue-on-error` skips records that fail to parse/evaluate (warns to stderr, exits non-zero)
- `tuple_filters` can't be expanded offline, so they're surfaced under `unresolved_filters`/stderr rather than dropped
- README documentation and integration test cases for `fga mapping run`

## Review Checklist
- [x] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev)

## Summary by CodeRabbit

* **New Features**
  * Added `fga mapping run` to evaluate JSON mapping events and emit tuple operations.
  * Supports JSONL streaming or JSON batch output, file or standard-input sources, and writes-only output for direct tuple importing.
  * Added aggregation to deduplicate results and detect conflicting write/delete operations.
  * Added continue-on-error handling with warnings for skipped records and unresolved filters.
* **Documentation**
  * Documented the new command, supported formats, options, and offline behavior.


rhamzeh (dismissed):

cmd/mapping/run.go
```diff
@@ -0,0 +1,522 @@
...
+	if event == nil {
+		return nil, nil, fmt.Errorf("reading input JSON: %w", errNonObjectRecord)
+	}
+
+	result, err := compiled.Evaluate(ctx, event)
+	if err != nil {
+		return nil, nil, fmt.Errorf("evaluating mapping: %w", err)
+	}
+
+	return result.Tuples, flattenFilters(result.TupleFilterOperations), nil
```
Siddhant-K-code: Filtered rules lose their desired tuples here. `mapper.Evaluate` stores them in `TupleFilterOperation.Tuples`, not `Result.Tuples`, so a valid `action: patch` rule returns empty `writes` and only the filter. Preserve each unresolved filter operation with its associated tuples, or return an error when the output format cannot represent it. Add a patch-filter regression test.

cmd/mapping/run.go
```diff
@@ -0,0 +1,522 @@
...
+			}
+
+			hadFailure = hadFailure || skipped
+
+			filters = append(filters, recFilters...)
+			buffered = append(buffered, tuples...)
+		}
+
+		if readErr != nil {
+			if errors.Is(readErr, io.EOF) {
```
Siddhant-K-code, line 342: This retains every unresolved filter until EOF in default JSONL mode, so memory grows with input size even though the command says this path streams. Emit filter warnings per record in streaming mode, and only accumulate filters for `--format json` or `--aggregate`.

ewanharris: Fixed by calling `warnUnresolvedFilters` immediately in processLine for the streaming path 

3b96f11ae3dc2701b5a2e169726b489bfb527a73

cmd/mapping/run.go
ewanharris: Fixed by threading []mapper.TupleFilterOperation through the whole pipeline instead of []language.TupleFilter, in  3b96f11ae3dc2701b5a2e169726b489bfb527a73

tests/mapping-test-cases.yaml
```diff
@@ -79,3 +79,106 @@ tests:
...
+    exit-code: 0
+    stdout:
+      line-count: 2
+
+  024 - run --aggregate dedups identical unresolved filters across records:
+    command: bash -c 'printf '"'"'{"id":"anne","org":"acme"}\n{"id":"anne","org":"acme"}\n'"'"' | fga mapping run --aggregate --format json ./tests/fixtures/filter-mapping.yaml'
+    exit-code: 0
+    stdout:
+      json:
+        unresolved_filters.#: "1"
```
Siddhant-K-code: This assertion still checks `unresolved_filters`, but the response field is now `tuple_filter_operations`. The `Tests` CI job fails because this path no longer exists. Update it to `tuple_filter_operations.#: "1"`.

sergiught (dismissed):

tests/mapping-test-cases.yaml
Siddhant-K-code:
```suggestion
        tuple_filter_operations.#: "1"
```

Siddhant-K-code (approved):

PR #763, opened 2026-09-17
feat(mapping): add interactive explorer to mapping run

ewanharris:
Adds an interactive explorer to `fga mapping run` via `-i`/`--interactive`. In a terminal it starts a REPL: type or paste a JSON document and see the tuple operations the mapping produces, rendered as an aligned table. It is aimed at authoring and debugging mappings without wiring up a store or a file of input records.

A document is evaluated as soon as it forms a complete JSON value. Continuation only happens where a newline is legal JSON whitespace, so an unterminated string or half-typed literal is evaluated and its error reported rather than silently waiting for input that can never make it valid. Invalid JSON is reported with the line, column, and a caret under the offending character.

A rule with `tuple_filters` cannot be resolved offline, so its filter conditions render as `filter` rows and the desired-state tuples it reconciles toward render as indented `desired` rows. These drive a read-diff-write against a store rather than being written directly (reconciliation can delete as well as write) so they are shown for inspection, not labelled as writes.

Demo below:

mapping-run-interactive

## Summary by CodeRabbit

- **New Features**
  - Added interactive mode to `fga mapping run` with JSON evaluation, multi-line input, trace controls, reload, quit commands, and formatted tuple results.
  - Added interactive mapping-file selection when no path is provided in supported mapping commands.
  - Added clearer diagnostics for invalid JSON, including line, column, and caret location.
  - Added filter and desired-state output for tuple filters.

- **Documentation**
  - Documented the interactive mode, supported commands, input behavior, and flag restrictions.


cmd/mapping/run_interactive.go
```diff
@@ -0,0 +1,535 @@
...
+	}
+
+	for _, filterOp := range ops {
+		for _, filter := range filterOp.Filters {
+			relation := filter.Relation
+			if relation == "" {
+				relation = "*"
+			}
+
+			fmt.Fprintf(writer, "  filter\t%s\t%s\t%s\n", filter.User, relation, filter.Object)
```
Siddhant-K-code: Preserve the filter semantics in this row. `TupleFilter.Action` distinguishes `patch` from `delete`, but both render as `filter`. Empty user and object fields also render blank even though they are wildcards. Render the action and normalize empty user, relation, and object fields to `*`.

cmd/mapping/run_interactive.go
```diff
@@ -0,0 +1,535 @@
...
+		}
+	}
+
+	_ = writer.Flush()
+}
+
+// tupleObject formats a tuple's object, appending its condition name in brackets
+// when the tuple is conditioned.
+func tupleObject(tuple language.Tuple) string {
+	if tuple.Condition != "" {
```
Siddhant-K-code: Include non-empty `tuple.Context` in the rendered tuple. Normal output includes both the condition and its context, but interactive mode shows only the condition name. Distinct conditional tuples can therefore appear identical.

cmd/mapping/run_interactive.go
```diff
@@ -0,0 +1,535 @@
...
+		fmt.Fprintf(s.errOut, "Error: %v\n", errNonObjectRecord)
+
+		return
+	}
+
+	result, err := s.compiled.Evaluate(s.ctx, event)
+	if err != nil {
+		fmt.Fprintf(s.errOut, "Error: evaluating mapping: %v\n", err)
+
+		return
```
Siddhant-K-code, line 429: Render `result.Trace` before returning this evaluation error. `mapper.Evaluate` returns a populated error trace with the error, but this return makes the `rules errored` output unreachable. With `:trace on`, evaluation failures currently show no rule trace.

cmd/mapping/run.go
```diff
@@ -495,9 +497,37 @@ before emitting; the default streaming JSONL does not.
...
+			if err := checkInteractiveFlags(runWritesOnly, runInputFile); err != nil {
+				fmt.Fprintln(errStream, "Error: "+err.Error())
+				os.Exit(2)
+			}
+		}
+
+		path := promptMappingFile(args, errStream)
+
+		if runInteractive {
+			if !isatty.IsTerminal(os.Stdin.Fd()) {
```
Siddhant-K-code: Validate that the interactive output is a TTY, or route the UI to `/dev/tty` or stderr. With `-i > out`, stdin passes this check, raw mode disables local echo, and the prompt, input echo, errors, and results go to the file. The user sees a blank terminal.

cmd/mapping/run_interactive.go
```diff
@@ -0,0 +1,535 @@
...
+	fileDescriptor := int(file.Fd())
+
+	oldState, err := term.MakeRaw(fileDescriptor)
+	if err != nil {
+		return fmt.Errorf("entering raw mode: %w", err)
+	}
+
+	defer func() { _ = term.Restore(fileDescriptor, oldState) }()
+
+	terminal := term.NewTerminal(readWriter{Reader: file, Writer: s.out}, promptPrimary)
```
Siddhant-K-code, line 187: Initialize `term.Terminal` with the actual TTY dimensions and update it on resize. `NewTerminal` defaults to 80x24, so cursor and repaint calculations are incorrect on terminals with a different width, corrupting edits of wrapped JSON lines.

cmd/mapping/run_interactive.go
ewanharris: Done in 4b45131. Rows now render the action (`filter:patch`/`filter:delete`) and unset user/relation/object show as `*`.

cmd/mapping/run_interactive.go
ewanharris: Done in 4b45131. `tupleObject` now appends the rendered context, e.g. `org:acme  [in_region {"region":"us"}]`

cmd/mapping/run_interactive.go
ewanharris: Done in 4b45131. `evaluate` now renders `result.Trace` before returning when :`trace on`

cmd/mapping/run.go
ewanharris: Done in 4b45131. `--interactive` now requires both stdin and stdout to be a TTY; a redirected stdout errors out instead of writing the UI to a file

cmd/mapping/run_interactive.go
ewanharris: Done in 4b45131. `runRaw` now seeds real dimensions via `term.GetSize` and tracks `SIGWINCH` resizes (no-op on Windows)

cmd/mapping/run_interactive.go
Siddhant-K-code: Verified in `4b45131`. Filter actions and all wildcard fields now render correctly.

cmd/mapping/run_interactive.go
Siddhant-K-code: Verified in `4b45131`. Conditional tuple context is now included in the output.

cmd/mapping/run_interactive.go
Siddhant-K-code: Verified in `4b45131`. Evaluation errors now include the available trace.

cmd/mapping/run.go
Siddhant-K-code: Verified in `4b45131`. Interactive mode now rejects redirected stdout.

cmd/mapping/run.go
```diff
@@ -495,9 +497,47 @@ before emitting; the default streaming JSONL does not.
...
+			}
+		}
+
+		path := promptMappingFile(args, errStream)
+
+		if runInteractive {
+			// Both streams must be a TTY: the explorer drives a raw-mode terminal,
+			// so a redirected stdout would send the prompt, echo, and results to a
+			// file and leave the user staring at a blank screen.
+			if !isatty.IsTerminal(os.Stdin.Fd()) || !isatty.IsTerminal(os.Stdout.Fd()) {
```
Siddhant-K-code, line 525: Move this TTY check before `promptMappingFile`. With `fga mapping run -i > out` and no mapping path, the command opens the file picker even though redirected stdout makes the invocation invalid. Validate the streams before any prompt so it fails immediately.

cmd/mapping/run_interactive.go
```diff
@@ -0,0 +1,581 @@
...
+// --writes-only and --format select machine output, --input reads from a file
+// instead of the prompt, and --aggregate and --continue-on-error act on a whole
+// input stream. Rejecting them is clearer than silently ignoring them.
+func checkInteractiveFlags(opts runMappingOptions, inputFile string) error {
+	switch {
+	case opts.writesOnly:
+		return errInteractiveWithWritesOnly
+	case inputFile != "":
+		return errInteractiveWithInput
+	case opts.format != "" && opts.format != "jsonl":
```
Siddhant-K-code: `--format jsonl` is accepted explicitly here, but `errInteractiveWithFormat` and the README say `--interactive` cannot be combined with `--format`. Either reject any explicitly supplied format using the flag’s `Changed` state, or update the error and docs to say only `--format json` is incompatible.

cmd/mapping/run_interactive.go
Siddhant-K-code: Verified initial sizing and Unix resize handling in `4b45131`. I am tracking the remaining Windows resize gap on the platform-specific no-op handler.

cmd/mapping/run_interactive_resize_windows.go
```diff
@@ -0,0 +1,27 @@
...
+limitations under the License.
+*/
+
+package mapping
+
+import "golang.org/x/term"
+
+// watchResize is a no-op on Windows, which has no SIGWINCH signal. The terminal
+// keeps the dimensions seeded at start-up.
+func watchResize(_ int, _ *term.Terminal) func() {
```
Siddhant-K-code: This leaves resize handling incomplete on Windows. `x/term.Terminal` does not refresh its dimensions automatically, so resizing the console leaves cursor and repaint calculations stale. Monitor console resize events or poll `term.GetSize` while reading input.

cmd/mapping/run_interactive.go
ewanharris: Done in b734ae4. Now rejects any explicit `--format` via the flag's `Changed` state (its default is non-empty `jsonl`), so `--format jsonl` is rejected too

cmd/mapping/run.go
ewanharris: Done in b734ae4. TTY and flag validation now run before `promptMappingFile`, so a redirected `-i > out` fails immediately instead of opening the picker

cmd/mapping/run_interactive_resize_windows.go
ewanharris: Done in b734ae4. Replaced the no-op with a 250ms poller: re-queries `term.GetSize` and pushes changes via `SetSize`. `SetSize`/`ReadLine` share an internal mutex so it's safe same as the Unix signal path

SoulPancake (dismissed):

cmd/mapping/run.go
```diff
@@ -495,9 +497,48 @@ before emitting; the default streaming JSONL does not.
...
+				fmt.Fprintln(errStream, "Error: --interactive requires an interactive terminal")
+				os.Exit(2)
+			}
+		}
+
+		path := promptMappingFile(args, errStream)
```
Siddhant-K-code, line 531: `promptMappingFile` renders to stderr, but the validation above checks only stdin and stdout. With `fga mapping run -i 2>log` and no mapping path, validation passes and the command blocks in an invisible picker. Require stderr to be a TTY when `len(args) == 0`, or render the picker through the validated terminal stream.

Siddhant-K-code (dismissed):
Overall looks good % comment

cmd/mapping/run.go
ewanharris: Changed `promptMappingFile` to require stdin and stderr to be a TTY also fixing the same in `validate` and `test` commands

sergiught (approved):

SoulPancake (approved):

PR #765, opened 2026-09-18
feat(docs): generate Commands section of README from cobra metadata

ewanharris:

The Commands section of README.md was hand-rolled and had drifted out of sync with the actual CLI — commands were missing, examples were outdated, and there was no mechanism to catch regressions.

#### How is it being solved?

A code generator (`cmd/gendocs`) walks the cobra command tree and splices a fresh Commands section and table-of-contents block into README.md. Command descriptions, parameters, examples, and response samples are sourced directly from each command's `Long`, `Example`, and `Annotations` fields, making the cobra definition the single source of truth.

A `make generate-docs` target runs the generator, and a new CI step fails the build if the committed README doesn't match the generated output, so drift is caught on every PR.

- `internal/doc`: new package containing the generator and splice logic, with tests
- `cmd/gendocs/main.go`: entrypoint that wires the root command into the generator
- `Makefile`: generate-docs target
- `.github/workflows/main.yaml`: Check README commands section is up to date CI step
- `Long`, `Example`, and `docs:response `/ `docs:response:lang` annotation fields added to all store, model, tuple, and query commands to populate the generated output
- `README.md`: Commands section and ToC block wrapped with splice markers; content regenerated

One change from this is that commands are now ordered alphabetically (cobra's `Commands()` order), so the generated section reads Authorization Models -> Manage JSON-to-tuple mappings -> Relationship Queries -> Stores -> Relationship Tuples rather than the previous hand-written order that kind of followed the natural flow of usage. Top-level runnables (version) are excluded from both the body and the ToC as they have no group heading.

## References

This is inspired by similar functionality in auth0-cli that is used to build the website for it.

## Review Checklist
- [x] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-reqests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).


rhamzeh (commented):
@ewanharris I like this idea, but is there any way we can keep the order of things in the readme?

internal/doc/generate.go
```diff
@@ -0,0 +1,202 @@
...
+				continue
+			}
+
+			writeSection(b, sub)
+		}
+
+		return
+	}
+
+	fmt.Fprintf(b, "##### %s\n\n", cmd.Short)
```
Siddhant-K-code, line 71: Render `cmd.Long` before the command block. The generated README never reads this field, so this PR deletes useful descriptions such as mapping interactive behavior and model test guidance even though the Cobra metadata contains them. Add a regression test that asserts `Long` appears in the generated section.

internal/doc/generate.go
```diff
@@ -0,0 +1,202 @@
...
+
+	fmt.Fprintf(b, "###### Command\n\n```\n%s\n```\n\n", cmd.UseLine())
```
Siddhant-K-code, line 73: Preserve required positional arguments in the generated syntax. `UseLine()` only includes arguments declared in `Use`; commands such as `query check`, `query expand`, and `tuple write` declare bare names and enforce operands through `Args`, so the README now shows incomplete commands like `fga query check`. Update those `Use` values or add explicit doc metadata, then test the generated syntax for an argument-taking command.

internal/doc/generate.go
```diff
@@ -0,0 +1,202 @@
...
+
+	if hasVisibleLocalFlags(cmd) {
+		fmt.Fprintf(b, "###### Parameters\n\n")
+
+		cmd.NonInheritedFlags().VisitAll(func(f *pflag.Flag) {
```
Siddhant-K-code, line 78: Include visible inherited flags here and in `hasVisibleLocalFlags`. Query commands define `--store-id`, `--model-id`, `--contextual-tuple`, `--context`, and `--consistency` on the parent, so their generated parameter lists omit required or important options. Iterate a flag set that includes inherited flags, while still filtering hidden flags, and add a parent persistent flag test.

Siddhant-K-code (commented):
The generator currently drops command descriptions, required positional arguments, and inherited flags. I left inline comments with specific fixes and regression tests.

PR #768, opened 2026-09-21
feat: support for dynamic conditions

aaguiarz:
## Description

This PR adds support for dynamic inline expression conditions in tuple YAML and CLI writes, including the $expression condition name and nested context.expression / context.parameters structure.

It also updates the example model and validation coverage to demonstrate a classic RBAC model combined with runtime-conditioned access checks.

## Review Checklist

## Summary by CodeRabbit

* **New Features**
  * Added optional `--condition-expression` and `--condition-parameters` flags to `fga tuple write` for defining dynamic, parameterized conditions.
  * Added validation to prevent incompatible flag combinations and invalid parameter usage.
  * Added an example demonstrating conditional access based on a channel name.
  * Added a dynamic RBAC example covering direct, role-based, and condition-based authorization.

* **Documentation**
  * Updated tuple-writing guidance with dynamic condition syntax and usage examples.


cmd/tuple/write.go
```diff
@@ -92,7 +93,18 @@ var writeCmd = &cobra.Command{
...
 
+const expressionConditionName = "$expression"
+
+var errExpressionConditionNotSupported = errors.New( //nolint:err113
+	"writing tuples with $expression conditions is not yet supported by the OpenFGA Go SDK; " +
+		"use 'fga model test' to test models that use inline expressions",
+)
+
 func writeTuplesFromArgs(cmd *cobra.Command, args []string, fgaClient *client.OpenFgaClient) error {
+	if cmd.Flags().Changed("condition-expression") {
```
Siddhant-K-code: This guard disables the advertised feature before `ParseTupleCondition` or the SDK call. The Go SDK already serializes `RelationshipCondition` unchanged, and OpenFGA v1.21 explicitly accepts `$expression` writes when `inline_expressions` is enabled. Please remove the argument, file, and store-import rejection guards, document the server feature requirement, and add an end-to-end write test.

cmd/tuple/write.go
```diff
@@ -295,6 +313,12 @@ func init() {
 	writeCmd.Flags().String("file", "", "Tuples file")
 	writeCmd.Flags().String("condition-name", "", "Condition Name")
 	writeCmd.Flags().String("condition-context", "", "Condition Context (as a JSON string)")
+	writeCmd.Flags().String("condition-expression", "", "Dynamic condition CEL expression")
+	writeCmd.Flags().String("condition-parameters", "", "Dynamic condition parameter types (as a JSON object)")
+	writeCmd.MarkFlagsMutuallyExclusive("condition-expression", "condition-name")
+	writeCmd.MarkFlagsMutuallyExclusive("condition-expression", "condition-context")
+	writeCmd.MarkFlagsMutuallyExclusive("condition-expression", "file")
+	writeCmd.MarkFlagsRequiredTogether("condition-expression", "condition-parameters")
```
Siddhant-K-code, line 304: `MarkFlagsRequiredTogether` makes `--condition-parameters` mandatory whenever `--condition-expression` is used. This contradicts the README, and OpenFGA accepts an expression without `parameters` by inferring referenced identifiers as strings. Please enforce only the one-way dependency: reject parameters without an expression, but allow an expression without parameters.

rhamzeh: @Siddhant-K-code where in the readme does it contradict this and what's the use-case for an expression with no parameters?

cmd/tuple/write.go
rhamzeh: Go SDK does not support expressions yet: https://github.com/openfga/go-sdk/blob/da52dbf7889cb44a8aff23ce25527ef357a71bd5/model_relationship_condition.go

cmd/tuple/write.go
aaguiarz: Actually, I did not test the tuple write command, and it might now work as it should require SDK support. Maybe we leave that out and only support it for 'model test'?

However, the model test worked, and that implies it should have been able to write the tuples..

Siddhant-K-code: @rhamzeh, README says params are optional & requ. an expression:

https://github.com/openfga/cli/blob/bd8689634891d59b9a50459021e94e8a4fcc3643/README.md#L774-L775

OpenFGA also has a passing test for `channel_id == 'X123456'` w/o declared params bcoz., undeclared identifiers are inferred as strings:

https://github.com/openfga/openfga/blob/ab557c5592670c899de35297e7aa067015f06502/internal/validation/validation_test.go#L1541-L1549

`MarkFlagsRequiredTogether` therefore adds the reverse req. Please reject parameters w/o an expression, but allow an expression w/o parameters

Siddhant-K-code: weird that GitHub didn't rendered the preview of second code block (maybe due to line 1500 thingy)

cmd/tuple/write.go
Siddhant-K-code: Agreed that the SDK has no expression-specific API. However, `$expression` fits the existing [`RelationshipCondition{Name, Context}`](https://github.com/openfga/go-sdk/blob/76d209a9753a5284db64df848eea41fea2506c6f/model_relationship_condition.go#L22-L26) payload. and, SDK fwd this through its normal write path, and [OpenFGA v1.21 accepts it when `inline_expressions` is enabled](https://github.com/openfga/openfga/blob/ab557c5592670c899de35297e7aa067015f06502/pkg/server/write.go#L58-L66).

Can we verify this w/ an SDK-backed write test before removing tuple-write support?

cmd/tuple/write.go
rhamzeh: Yes, they are optional - as in they are not required, and if provided they will require the expression.

rhamzeh: @senojj can I ask you to jump in here as you implemented it on the API.

Do you want us to expose the ability to set expressions with no parameters?

I lean towards no unless there's a very good use case for it.

aaguiarz: I agree, let's not allow it, I'm changing it

rhamzeh (approved):

